PR #18, opened 2015-02-15
Fix UUID formatting (returned as a string)

tck42:
Per rfc4530 2.1 UUID (1.3.6.1.1.16.1) syntax is just an ascii string with the UUID's representation instead of its bytes (even though the RFC describes the underlying data format in the first paragraph). Sending this string to UUID(bytes=) results in a ValueException (to recreate, request entryuuid from openldap which uses this syntax).

### DIFF
--- a/ldap3/protocol/formatters/standard.py
+++ b/ldap3/protocol/formatters/standard.py
@@ -92,7 +92,7 @@ standard_formatter = {
     '1.3.6.1.4.1.1466.115.121.1.56': format_binary,  # LDAP Schema Definition [OBSOLETE]
     '1.3.6.1.4.1.1466.115.121.1.57': format_unicode,  # LDAP Schema Description [OBSOLETE]
     '1.3.6.1.4.1.1466.115.121.1.58': format_unicode,  # Substring assertion
-    '1.3.6.1.1.16.1': format_uuid,  # UUID
+    '1.3.6.1.1.16.1': format_unicode,  # UUID
     '2.16.840.1.113719.1.1.4.1.501': format_uuid,  # GUID (Novell)
     '2.16.840.1.113719.1.1.5.1.0': format_binary,  # Unknown (Novell)
     '2.16.840.1.113719.1.1.5.1.6': format_unicode,  # Case Ignore List (Novell)


cannatag:
Hello tck42,
rfc 4530 states that:
   In LDAP, UUID values are encoded using the [ASCII] character string
   representation described in [RFC4122].  For example,
   "597ae2f6-16a6-1027-98f4-d28b5365dc14".

that's how the formatter for 1.3.6.1.1.16.1 works, it get a byte sequence and return a formatted UUID string. If you want the underlying bytes you can read the 'raw_attributes' dictionary (for example c.response[0]['raw_attributes']['UUID'] or c.entries[0].GUID.raw:values if c is your Connection ojbect).

Also you get the formatted output only if you read the schema and set in the connection the 'check_names' parameter to True. In this case you get the formatted attribute in the 'attributes' dictionary and the raw values in the 'raw_attributes' dictionary of the response.

Most formatter in ldap3 are not reversable, they are used to have more meaningful attributes values.

What do you think?
Giovanni

tck42:
I think what you're saying is what I was expecting too, but after testing I don't think it's the case (at least with openldap).
The problem is the byte sequence you get back (at least from my testing with openldap) isn't the 16-byte octetstring (that would be suitable for passing to the bytes keyword in the UUID constructor) - the conversion to ascii string representation happens before the value is sent from the ldap server to the client, so you get the ascii encoded printable representation. I think it's easier to show by example:
tck42@gon:~/work/ldap3
$ ldapsearch -LLL -x entryuuid
dn: dc=tck42,dc=com
entryUUID: aaa4c234-8659-46da-89ca-308a2785865b

The following is the output from searching via ldap3 (with the patch applied, and no other differences from your master branch). Here are the two lines responsible for output:
    print(r['dn'], r['attributes']) # return unicode attributes
    print(r['dn'], r['raw_attributes']) # return raw (bytes) attributes
And here's the output:
dc=tck42,dc=com {'entryUUID': 'aaa4c234-8659-46da-89ca-308a2785865b'}
dc=tck42,dc=com {'entryUUID': [b'aaa4c234-8659-46da-89ca-308a2785865b']}

Note that raw attributes _doesn't_ have b"\xaa\xa4\xc24\x86YF\xda\x89\xca0\x8a'\x85\x86[" which is what I expected initially, and I believe what your code expects (and requires to work properly). Without the patch applied I get the following:
  File "test2.py", line 5, in <module>
    c.search('dc=tck42,dc=com','(objectClass=*)', BASE, attributes=['entryuuid'])
  File "/home/tck42/.virtualenvs/ldap3/lib/python3.4/site-packages/ldap3/core/connection.py", line 483, in search
    response = self.post_send_search(self.send('searchRequest', request, controls))
  File "/home/tck42/.virtualenvs/ldap3/lib/python3.4/site-packages/ldap3/strategy/sync.py", line 131, in post_send_search
    responses, result = self.get_response(message_id)
  File "/home/tck42/.virtualenvs/ldap3/lib/python3.4/site-packages/ldap3/strategy/base.py", line 265, in get_response
    responses = self._get_response(message_id)
  File "/home/tck42/.virtualenvs/ldap3/lib/python3.4/site-packages/ldap3/strategy/sync.py", line 155, in _get_response
    dict_response = self.decode_response(ldap_resp)
  File "/home/tck42/.virtualenvs/ldap3/lib/python3.4/site-packages/ldap3/strategy/base.py", line 363, in decode_response
    result = search_result_entry_response_to_dict(component, self.connection.server.schema, self.connection.server.custom_formatter, self.connection.check_names)
  File "/home/tck42/.virtualenvs/ldap3/lib/python3.4/site-packages/ldap3/operation/search.py", line 472, in search_result_entry_response_to_dict
    entry['attributes'] = checked_attributes_to_dict(response['attributes'], schema, custom_formatter)
  File "/home/tck42/.virtualenvs/ldap3/lib/python3.4/site-packages/ldap3/operation/search.py", line 401, in checked_attributes_to_dict
    checked_attributes[str(attribute['type'])] = format_attribute_values(schema, str(attribute['type']), decode_raw_vals(attribute['vals']) or [], custom_formatter)
  File "/home/tck42/.virtualenvs/ldap3/lib/python3.4/site-packages/ldap3/protocol/formatters/standard.py", line 185, in format_attribute_values
    formatted_values = [formatter(raw_value) for raw_value in values]
  File "/home/tck42/.virtualenvs/ldap3/lib/python3.4/site-packages/ldap3/protocol/formatters/standard.py", line 185, in <listcomp>
    formatted_values = [formatter(raw_value) for raw_value in values]
  File "/home/tck42/.virtualenvs/ldap3/lib/python3.4/site-packages/ldap3/protocol/formatters/formatters.py", line 63, in format_uuid
    return str(UUID(bytes=raw_value))
  File "/usr/lib64/python3.4/uuid.py", line 148, in __init__
    raise ValueError('bytes is not a 16-char string')
ValueError: bytes is not a 16-char string

This is a result of passing the ascii-encoded representation to the bytes keyword on the UUID constructor. If you just passed it as the main argument it would work, but that'd be pointless extra work since you just convert back to ASCII, in the process creating a UUID for nothing.
I don't think the formatter is correct for that OID. For what it's worth I think the rfc is written in an extremely confusing fashion - they describe the internal UUID format in some detail and then state "ldap sends these as the ASCII encoded representation"; I'm curious if the above error comes through for any other server implementations with that OID. I agree that disabling pulling the schema also fixes, but that is a workaround that should not be necessary, and disables formatting on the other attributes that work properly.

Also please see this response too:
http://www.openldap.org/lists/openldap-devel/200908/msg00050.html

So that OID (strangely, in my opinion) sends an LDAP String to the client.

Full test code (basically your tutorial code but pulling back entryuuid instead) - if you have an openldap available just update the base dn to recreate:
from ldap3 import Connection, Server, ANONYMOUS, SIMPLE, SYNC, ASYNC, ALL, SUBTREE, BASE, LEVEL, NONE
s=Server('127.0.0.1',port=389,get_info=ALL)
c=Connection(s, auto_bind=True, client_strategy=SYNC)
c.search('dc=tck42,dc=com','(objectClass=*)', BASE, attributes=['entryuuid'])
response=c.result
for r in response:
    print(r['dn'], r['attributes']) # return unicode attributes
    print(r['dn'], r['raw_attributes']) # return raw (bytes) attributes
c.unbind()
Thanks!

cannatag:
Thanks for your clarifying message. I think it would be better to modify the format_uuid() function to check if the argument is a 16 bytes long binary and only in that case perform the conversion to string, else it try to unicode() it. It neither of this is valid it should return the original argument umodified.
Later I will provide a patch for you to try.

Giovanni

tck42:
Thanks! I agree this is the safest approach.

cannatag:
Please change the definition of format_uuid() in ldap3.protocol.formatters.formatters.py at line 61 with:

    def format_uuid(raw_value):
        try:
            return str(UUID(bytes=raw_value))
        except (TypeError, ValueError):
            return format_unicode(raw_value)
        except Exception:
            pass

        return raw_value


and tell me if this works as expected.

bye,
Giovanni

tck42:
Sorry for the delay - yep, this works fine. Thanks!

cannatag:
ok. I will include in upcoming 0.9.7.5.

Bye,
Giovanni

PR #29, opened 2015-02-27
Include `_version.json` file in source distribution

cgohlke:
`ldap3-0.9.7.9.tar.gz` on PyPI is missing the `_version.json` file. Hence setup.py fails at https://github.com/cannatag/ldap3/blob/master/setup.py#L31


cannatag:
Thanks, I had a lot of problems with pypi lately. I backrev to 0.9.7.5 and fix the source distribution.

Sorry for the inconvenience.

Giovanni

PR #34, opened 2015-03-31
The default port for cleartext LDAP is 389

Metuchen:


cannatag:
What a mistake! Thanks I will modify docs immediately.

PR #36, opened 2015-04-04
Add raw SASL Credentials field.

pefoley2:
When using python3, the conversion to str mangles the saslCreds
response.
Add a field to access the raw bytes, this is required for GSSAPI.


pefoley2:
I found a few issues with this, I'll re-submit soon.

PR #37, opened 2015-04-05
Use bytes instead of str for SASL Credentials field.

pefoley2:
When using python3, the conversion to str mangles the saslCreds response thus breaking GSSAPI.
Convert the saslCreds field to bytes, which fixes GSSAPI.


cannatag:
THanks Peter, I've slightly modify your code. Can you check if dev works for you?

pefoley2:
It works, thanks!

PR #38, opened 2015-04-05
fix paged search while using the Reader.search_paged() method.

wazboy:
I was not able to get paged searches working with the Reader abstraction.  I changed the behavior of the method.

* The cookie was not saved in Reader._execute_query().
  *  This results in only getting the first page of query results on each invocation.
* The Reader.search_paged() method would only yield once.
* The Reader.search_paged() method would yield a list of entries, not each entry found.

This pull request modifes the Reader._execute_query() method to save the cookie.  It also changes the ergonomics of Reader.search_paged() so it can be used as follows:

```python
    person_reader = Reader(c, person_def, '', 'ou=Users,dc=lazarus,dc=com')

    for person in person_reader.search_paged(5):
        print(person.givenName, person.sn)
```

### DIFF
--- a/ldap3/abstract/reader.py
+++ b/ldap3/abstract/reader.py
@@ -376,6 +376,13 @@ class Reader(object):
                                             paged_size=self.paged_size,
                                             paged_criticality=self.paged_criticality,
                                             paged_cookie=self.paged_cookie)
+
+            result = self.connection.result
+            try:
+                self.paged_cookie = result['controls']['1.2.840.113556.1.4.319']['value']['cookie']
+            except KeyError:
+                self.paged_cookie = None
+
             if not self.connection.strategy.sync:
                 response, _ = self.connection.get_response(result)
             else:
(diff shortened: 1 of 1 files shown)


cannatag:
Thanks, I will test your code and include it in the next release of ldap3-

Bye,
Giovanni

cannatag:
Hi wazboy,
I've removed the paged_search logic from the reader object and have integrated the newer connection.extend.paged_search() method in the reader itself. I prefer this approach because I have the paged search logic in only one place. With the parameter generator=True|False You can choose to have the reader to behave as a generator (ldap searches are performed while you consume the entries) or accumulator (all paged search are performed before you read the first entry). I've committed the changes in dev. Can you test if this is working for you?

Thanks, 
Bye,

wazboy:
Hi cannatag,

I've tested the changes in the dev branch.  This has the same semantica behavior as the changes in this pull request.  As such it works perfectly.  I agree that leveraging the existing paged_search from the extended module promotes code reuse.  You changes are superior to this pull request.

Regards

cannatag:
Thanks for testing. I will release it in 0.9.8.2

Bye,
Giovanni

PR #48, opened 2015-05-01
Fix empty message type in connection usage

### LINKED ISSUE #47: ConnectionUsage.update_transmitted_message() crash on bind request
Hi,

I just noticed that when the bind occurs, the message['type'] is empty and makes the method update_transmitted_message() crash.

How to reproduce:
Try to bind with `Connection(..., collect_usage=True, ...)`.
Here is how I init my connection object:
```
self.connection = Connection(self.server,
                                 auto_bind=True,
                                 client_strategy=SYNC,
                                 user=username,
                                 password=password,
                                 authentication=SIMPLE,
                                 check_names=True,
                                 collect_usage=True)
```

It would crash in usage.py on line 161, at the first if:

```
 def update_transmitted_message(self, message, length):
        self.bytes_transmitted += length
        self.operations += 1
        self.messages_transmitted += 1
        if message['type'] == 'abandonRequest':      <---- crashes because message is None
            self.abandon_operations += 1
...
```

It is possible to quick fix by putting the IFs in a try/catch block, but it doesn't fix the issue that message['type'] isn't set for bind requests.

UltraPhil:
Fixes #47

### DIFF
--- a/ldap3/core/usage.py
+++ b/ldap3/core/usage.py
@@ -158,28 +158,31 @@ class ConnectionUsage(object):
         self.bytes_transmitted += length
         self.operations += 1
         self.messages_transmitted += 1
-        if message['type'] == 'abandonRequest':
-            self.abandon_operations += 1
-        elif message['type'] == 'addRequest':
-            self.add_operations += 1
-        elif message['type'] == 'bindRequest':
-            self.bind_operations += 1
-        elif message['type'] == 'compareRequest':
-            self.compare_operations += 1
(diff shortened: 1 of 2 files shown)


UltraPhil:
Sorry if I'm not doing it properly, it's the first time I attempt to do a pull request.

cannatag:
Thanks, I will check it soon.

Bye,
Giovanni

UltraPhil:
I will re-do pull request with good commit.

PR #49, opened 2015-05-01
Fix the None object problem

UltraPhil:
Addresses issue #47


cannatag:
Thanks,
will include in next release of ldap3. 

Bye,
Giovanni

PR #66, opened 2015-06-16
Corrected remaining references to NETWORK_EXTENDED

acaso:
Corrected remaining references to NETWORK_EXTENDED instead of EXTENDED in logging.


cannatag:
Fixed, thanks. It should be available on ldap3.readthedocs.org.

Bye,
Giovanni

PR #82, opened 2015-07-09
Unbundle backported tsl code

mcyprian:
ldap3 currrently bundles it's own backport of Pyhton's ssl.match_hostname.
In this PR, the backport from [backports.ssl_match_hostname](https://pypi.python.org/pypi/backports.ssl_match_hostname) is used instead.
The benefit of this is that there is no need to keep track of any security risks and fixes of the given code.
More commits will follow with adding this to setup.py's install_requires.


cannatag:
Hi Michal,
I'd prefer to have the same package requirements (actually only pyasn1) because I switch frequently from python2 to python3 boxes. It seems simpler to me to have the same codebase working seamlessly in both version without any additional package required. I could explain in the docs that for python before 2.7.10 the library can take advantage of the backports.ssl_match_hostname module if present.

cannatag:
Please check 0.9.8.7. It uses back ports.check_hostname if present. I have also updated the back ported code in utile.tls_backport to the code in 3.4.3. 

Could this be sufficient for you? By the way I peeked the 3.5 code and it includes check of ip host using the ipaddress module, not present in python2. So any new backport should include the ipaddress backport too.

Bye,
Giovanni

mcyprian:
This solution is acceptable, thank you for the notice.

Michal

PR #86, opened 2015-07-22
Avoiding TypeError and Ordered object class

yykamei:
I fixed add method in ldap3.Connection.

parm_object_class and attr_object_class should be same type.
And I propose objectClass attributes remember its order.

### DIFF
--- a/ldap3/core/connection.py
+++ b/ldap3/core/connection.py
@@ -676,14 +676,14 @@ class Connection(object):
             if object_class is None:
                 parm_object_class = []
             else:
-                parm_object_class = object_class if isinstance(object_class, SEQUENCE_TYPES) else [object_class]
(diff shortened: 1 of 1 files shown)


cannatag:
Thanks Yutaka, I will check your code and include it in the next release of ldap3.
Can I ask why you need ordered attributes in LDAP? This should not be an issue with the protocol.

Bye,
Giovanni

yykamei:
Because unfortunately some old application require ordered attributes.
Of course, LDAP does not require ordered attributes and
LDAP client MUST not expect ordered  attributes.

You can ignore it because it is just a proposal :-)

Thanks,

cannatag:
Hi Yutaka, I will include the fix and the ordered objectclass values list in the next release 0.9.8.8 of ldap3.

PR #87, opened 2015-07-24
fixed evaluate_match to handle assertion values containing filter types

eehakkin:
According to RFC4515 section 3, an LDAP search filter

    (cn=a<=b=>c)

represents an equal match having an attribute "cn" and an assertion value "a<=b=>c" instead of representing a less or equal match having an attribute "cn=a" and and assertion value "b=>c".

This patch fixes ldap3.operation.search.evaluate_match to handle that and similar cases correctly.

### DIFF
--- a/ldap3/operation/search.py
+++ b/ldap3/operation/search.py
@@ -103,29 +103,27 @@ class FilterNode(object):
 
 
 def evaluate_match(match, schema):
-    match = match.strip()
-    if '~=' in match:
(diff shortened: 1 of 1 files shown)


cannatag:
thanks, I will check your fix soon.

Bye

cannatag:
Thanks Eero,
I've merged your fix and added the test_parse_search_filter_equality_2() test in testParseSearchFilter to check this case. I will include this in the next release of ldap3.

Bye,
Giovanni
:

PR #89, opened 2015-08-03
Support round-robin DNS with GSSAPI auth

alexsdutton:
Our university has multiple LDAP servers presented in a round-robin when one resolves the domain name for our LDAP service. Each of those servers has its own service principle, something like `ldap-<n>.oak.ox.ac.uk`.

Previously, the GSSAPI negotiation expected to authenticate against the round-robin DNS name, not the hostname of the server it connected to. This pull request performs an rDNS on the server we're connected to and then attempts to negotiate with that name.

### DIFF
--- a/ldap3/protocol/sasl/kerberos.py
+++ b/ldap3/protocol/sasl/kerberos.py
@@ -26,6 +26,8 @@
 # original code by Hugh Cole-Baker, modified by Peter Foley
 # it needs the gssapi package
 
+import socket
+
 from ...core.exceptions import LDAPPackageUnavailableError, LDAPCommunicationError
 
 try:
@@ -47,7 +49,8 @@ def sasl_gssapi(connection, controls):
     from RFC 4752. Does not support any security layers, only authentication!
     """
 
-    target_name = gssapi.Name('ldap@' + connection.server.host, gssapi.NameType.hostbased_service)
+    hostname = socket.gethostbyaddr(connection.socket.getpeername()[0])[0]
+    target_name = gssapi.Name('ldap@' + hostname, gssapi.NameType.hostbased_service)
     ctx = gssapi.SecurityContext(name=target_name, mech=gssapi.MechType.kerberos)
     in_token = None
     try:


alexsdutton:
I'm not sure that would work, as there's no way to know which host you'll be connecting to ahead of time. Would it be reasonable to make sasl_credentials a dict, and look for an `rdns` key that would mirror the functionality in http://web.mit.edu/kerberos/krb5-1.12/doc/admin/princ_dns.html#service-principal-canonicalization? Perhaps default to `False` (no canonicalization) if it's absent, to maintain the current behaviour?

alexsdutton:
And now I've realised that `sasl_credentials` is a tuple in other bits of the code. Hmmm.

alexsdutton:
http://web.mit.edu/kerberos/krb5-1.12/doc/appdev/gssapi.html says that the rDNS canonicalisation should happen automatically. Let me investigate further…

cannatag:
Hi, I've changed the code, if the first element of sasl_credentials is True the target_name will be requested from DNS, if it contains a string it will be used as target name, if it's empty of False (or no sasl_credentials attribute is specified) the connection hostname will be used:

```python
f connection.sasl_credentials and len(connection.sasl_credentials) == 1 and connection.sasl_credentials[0]:
        if connection.sasl_credentials[0] is True:
            hostname = socket.gethostbyaddr(connection.socket.getpeername()[0])[0]
            target_name = gssapi.Name('ldap@' + hostname, gssapi.NameType.hostbased_service)
        else:
            target_name = gssapi.Name('ldap@' + connection.sasl_credentials[0], gssapi.NameType.hostbased_service)
    else:
        target_name = gssapi.Name('ldap@' + connection.server.host, gssapi.NameType.hostbased_service)
    creds = gssapi.Credentials(name=gssapi.Name(connection.user), usage='initiate') if connection.user else None
    ctx = gssapi.SecurityContext(name=target_name, mech=gssapi.MechType.kerberos, creds=creds)
```

Can you check this in dev?

Bye,
Giovanni

alexsdutton:
That's great; thank you. Would you like me to write some documentation for it?

cannatag:
Yes, please do! I would really appreciate it and any other information related to kerberos authentication because I don't use it at all.

Bye,
Giovanni

PR #90, opened 2015-08-03
Allow choosing credentials with GSSAPI

alexsdutton:
If an application is using a credential cache collection with multiple credentials, it would be handy to have a way to specify which credentials to use in the GSSAPI negotiation.

This reuses `connection.user` to find the desired credentials.


cannatag:
I will merge in dev.

PR #97, opened 2015-08-08
Docs improvements for GSSAPI auth

alexsdutton:
Here is some documentation for the changes made in 3fa34ce, as offered in #89.

I've also added some links to the bind and search documentation from other bits of the docs (quicktour and autogen'd docs for `ldap3.protocol.sasl.kerberos`).


cannatag:
merged. thanks a lot for the docs!

PR #103, opened 2015-08-31
In later gssapi, ctx.complete no longer raises an exception if incomplete.

alexsdutton:
Raising an exception was a bug which has been fixed in a recent gssapi (pythongssapi/python-gssapi#70). Now handle both an exception, and `ctx.complete` being `False`.


cannatag:
Thanks, I will include your fix in the next release.

PR #104, opened 2015-09-03
Added response_to_dict_list() method to the Connection object.

w-p:
When returning entries, the resultant collection can require additional processing on the client or server side to extract the attribute representation of each entry. This produces a list of entries in plain dict() format such that each entry is simply {attribute: value,...}.

### DIFF
--- a/ldap3/core/connection.py
+++ b/ldap3/core/connection.py
@@ -1096,6 +1096,26 @@ class Connection(object):
 
             return None
 
+    def response_to_dict_list(self,
+                         raw=False,
+                         search_result=None):
+        with self.lock:
+            if search_result is None:
+                search_result = self.response
+
+            if isinstance(search_result, SEQUENCE_TYPES):
+                dict_list = list()
+
+                for response in search_result:
+                    if response['type'] == 'searchResEntry':
+                        if raw:
+                            entry = dict(response['raw_attributes'])
+                        else:
+                            entry = dict(response['attributes'])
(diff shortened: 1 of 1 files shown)


cannatag:
Hi Will, why did you close this request? It's not needed anymore?

Bye,
Giovanni

w-p:
A plain, formatted dictionary representation of entries is needed but my forked implementation was deeply faulted as I didn't handle attribute formatting so I retracted the pull request.

I have a work around for now but I'm not sure how to standardize this for use in ldap3.  The work around is essentially:
`get_attributes = lambda x: x.get('attributes')`
`entries = json.loads(connection.response_to_json())`
`entries = list(map(get_attributes, entries.get('entries')))`

This is clearly a hack.  Worse, the next step is to convert the list of dict objects back into JSON.

cannatag:
Hi Will, 
you mean a dictionary where the key is the dn of the returned object and the value is another dictionary with the attributes values:

entries =  {r['dn']: r['attributes'] for r in c.response if r['type'] == 'searchResEntry'}

or you prefer a list of tuples where each tuple is a response with its first value set to the dn of the returned entry and the latter the attributes values:

entries = [(r['dn']: r['attributes']) for r in c.response if r['type'] == 'searchResEntry']

The former solution has the advantage to be indexed by the object dn, while the latter is easily iterable. I'd prefer the latter because usually you don't know in advance the dn of the objects returned by a search.

Bye,
Giovanni

PR #123, opened 2015-11-14
Add escape_filter_chars function

pefoley2:
Allows for properly escaping arbitrary input to properly conform to
RFC4515.

Based on ldap.filter.escape_filter_chars from the python-ldap library.


cannatag:
thanks, will include it in the next release of ldap3.

PR #139, opened 2015-12-09
Unbreak set_config_parameter function

sigmunau:


cannatag:
thanks, I messed up with cut and paste... I will include it in the next release of ldap3.

Bye,
Giovanni

PR #141, opened 2015-12-11
Make contains work with false values

Vampouille:
I think contains will reply false if it contains a false value. But It should return true. This last commit fix this issue.


cannatag:
Thanks, you're right. I will include it in the next release.

PR #150, opened 2016-02-16
Fix __getattr__ not throwing AttributeError (did throw KeyError)

kilroy42:
hasattr(entry, 'a-string') failed in the old version - because __getattr__ MUST throw a AttributeError and __getitem__ must throw a KeyError.


cannatag:
thanks. Will include it in the next release of ldap3.

Bye,
Giovanni

PR #162, opened 2016-04-17
Added LM:NTLM hash auth support to NTLM bind

dirkjanm:
I'm not sure if you are interested in a feature like this, but as a pentester it is often useful to authenticate to the server with a NTLM hash rather than a password.
I've added the option to specify this as a LM:NTLM hash (similar to how metasploit does it), which works when done as in the added documentation option (that are the hashes for "password").

I haven't tested building the docs, so I hope my formatting is correct there.


cannatag:
Thanks, will include in next release of ldap3.

Bye,
Giovanni

PR #170, opened 2016-04-28
Remove add_user and remove_user methods in mockSync strategy

roxanagherle:
Mock users can be seen as regular mock entries, therefore add_entry and remove_entry can be used instead.


cannatag:
thanks again.

bye.

PR #175, opened 2016-05-09
setup.py to read from requirements.txt

joedborg:
This means you don't have to keep 2 files up-to-date.


cannatag:
thanks. really a nice trick!

Giovanni

PR #191, opened 2016-07-01
Update docs: use_tls --> use_ssl

brookskindle:
Noticed a small typo when reading through the docs. There is no `use_tls` parameter for a `Server` object, it's `use_ssl`.


cannatag:
thanks! will fix it immediately.

PR #197, opened 2016-07-14
Escape all octets greater than 0x7F that are not part of a valid UTF-8.

cfelder:
In respect of RFC 4515 these characters must be escaped too.


cannatag:
Released in 1.4.0.

PR #200, opened 2016-07-17
Only use default certs if no custom CA info given

ahnolds:
Only use default certs if no custom CA info given.  This matches the behavior of `ssl.create_default_context` [from the builtin ssl module](https://github.com/python/cpython/blob/78644cf0f4b475328c3b59e4d73c84df8a2ab7aa/Lib/ssl.py#L471).


cannatag:
Thanks. I will check and include it in the next release of ldap3.

cannatag:
Released in 1.4.0.

PR #206, opened 2016-08-02
Remove logging.info() line.

dsully:
Modules should not emit logging info without being asked.


cannatag:
You're right. I will include in the next realease of ldap3.

PR #217, opened 2016-09-21
Change response_to_json to response_to_file in mocking.rst

RamblingCookieMonster:
Hi!

The example calls `response_to_json` with a file to output - this doesn't appear to work ([see method args](https://github.com/cannatag/ldap3/blob/master/ldap3/core/connection.py#L1181)).

Updated the example to call `response_to_file` instead, [which appears to call](https://github.com/cannatag/ldap3/blob/master/ldap3/core/connection.py#L1233) `response_to_json` on it's own.

Feel free to ignore this, easy to work around, and there may be more downstream changes that I've missed.  Figured it might save someone some digging.

Cheers!


cannatag:
thanks. Will check and include in the next release of ldap3.

cannatag:
thanks. docs updated!

PR #234, opened 2016-10-28
Fix syntax in changelog, fixing rendering as well.

jaraco:


cannatag:
thanks, manually fixed in 2.0.9.

PR #241, opened 2016-11-02
Allow searches of SIDs as a strings on mock server

mprahl:
I am writing unit tests for my code that uses ldap3 on Python 2.7, but when searching my mocked LDAP server with the filter `(&(objectClass=group)(objectSid=S-1-5-21-1270288957-3800934213-3019856503-513)`, I don't get any results back even though there is a group with an objectSid of that value.

Since the searches are going against the raw attributes, the value of objectSid is in binary format, so there isn't a match. This pull request adds the ability to search by a string version of an objectSid on a mock LDAP server.

Feel free to change the PR as you see fit.


cannatag:
Please check code in dev. It now applies standard formatters to all stored attributes, not only to objectSid.

mprahl:
@cannatag thanks! I'll try it tonight.

mprahl:
@cannatag it worked. Thank you.

Issue #250, opened 2016-11-09
AttributeError is raised if no stdin is present


osemenovsky:
When running script using ldap3 lib without a tty, an AttributeError exception is raised

cannatag:
Updated in dev. Can you check it? Thanks

cannatag:
will be released in 2.1.1

cannatag:
Released in 2.1.1

osemenovsky:
Awesome, thanks

PR #251, opened 2016-11-09
fixed AttributeError in case of stdin absence

osemenovsky:
modified:   config.py

https://github.com/cannatag/ldap3/issues/250


cannatag:
Thanks. Will include it in the next version of ldap3.

PR #262, opened 2016-11-21
added support for finding object by GUID on AD in safe_dn()

dinhngtu:
MS Active Directory lets you use DN syntaxes like '<GUID=7b95f0d5-a3ed-486c-919c-077b8c9731f2>' to look up objects by GUID.

This PR modifies DN sanitization code in safe_dn() to support this DN syntax.


cannatag:
ok, will include in the next release. thanks

cannatag:
Included in dev. thanks.

PR #277, opened 2016-12-28
fix line breaks in the enumeration

cornelinux:
The line breaks within the bullet points are a bit confusing. Fix the indentation so that all information is under one bullet point.


cannatag:
Thanks, will fix in dev.

PR #281, opened 2017-01-08
MockSyncStrategy: Implement request decoding

fabianfreyer:
**Alternative to #282: Please only merge one of these PRs**

Until now, `MockSyncStrategy` did not save requests in the `connection`
object. This breaks the automatic Reader generation in
`connection._get_entries`, rendering connection.entries unusable.

This patch implements a `decode_request` staticmethod for the
MockBaseStrategy analogous to the `BaseStrategy`, from where the code
is mostly copied over. I'd put it up to discussion if this code duplication is
the right way to go, or whether the `BaseStrategy.decode_request` should
have the `message_type` and `component` extraction removed, so that it
can also be used by `Mock{Base,Sync}Strategy`.

### DIFF
--- a/ldap3/strategy/mockBase.py
+++ b/ldap3/strategy/mockBase.py
@@ -617,3 +617,32 @@ class MockBaseStrategy(object):
             pass
 
         return False
+
+    @staticmethod
+    def decode_request(message_type, component):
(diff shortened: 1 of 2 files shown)


cannatag:
I've included #282 in dev.

PR #282, opened 2017-01-08
MockSyncStrategy: Implement request decoding (alternative)

fabianfreyer:
**Alternative to #281: Please only merge one of these PRs**

Until now, MockSyncStrategy did not save requests in the connection
object. This breaks the automatic Reader generation in
connection._get_entries, rendering connection.entries unusable.

This patch removes some seemingly redundant logic that separating the
message type and request from the ldap message in the
BaseStrategy.decode_request staticmethod. Both of these values are
already known by the callers, allowing it to be used in the
MockSyncStrategy.

### DIFF
--- a/ldap3/strategy/base.py
+++ b/ldap3/strategy/base.py
@@ -294,7 +294,7 @@ class BaseStrategy(object):
             message_controls = build_controls_list(controls)
             if message_controls is not None:
                 ldap_message['controls'] = message_controls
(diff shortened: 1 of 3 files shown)


cannatag:
Thanks, I've included this in dev. Prefer not to merge with master..

Can you give a try? Dev includes many minor fixes too.

cannatag:
done in v2.2.0

PR #293, opened 2017-01-27
Fix ObjectDef remove_attribute

mattcl:
Properly handle when item is `AttrDef`:
`item == attr.lower()` is an `AttrDef` to `basestring` comparison, which will
always return `False` because of the implementation of `__eq__` on `AttrDef`.
Replace with `key == attr.lower()`

This did not happen in the following because the implementation of `__isub__` does an `isinstance` check and passes `other.key` in the case where `other` is `AttrDef`
```
o = ObjectDef(...)
o -= 'foo'
```

`LDAPAttributeError` was was returning type of key instead of item (key
would always be `NoneType` in the situation that item was not string or
`AttrDef`)

I would have added a test for this, but I don't see a test case that tests `ObjectDef` directly.

### DIFF
--- a/ldap3/abstract/objectDef.py
+++ b/ldap3/abstract/objectDef.py
@@ -195,17 +195,17 @@ class ObjectDef(object):
         if isinstance(item, STRING_TYPES):
             key = ''.join(item.split()).lower()
         elif isinstance(item, AttrDef):
-            key = item.key
+            key = item.key.lower()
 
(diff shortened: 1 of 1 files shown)


cannatag:
Thanks, will include it in the next release of ldap3.

Bye,
Giovanni

cannatag:
thanks

Issue #301, opened 2017-02-21
search result decode other than utf-8


vaclavrak:
Hi, recently I found problem in our LDAP implementation search result responses are not in UTF-8, but windows-1250. In current code  is hardcoded utf-8 (operations/search.py:524)   

```Python
entry_dict['dn'] = response[0][3].decode('utf-8')  
```

in my example the response[0][3] value is like 
```Python 
> b'CN=Rak V\xe1clav'.decode('utf-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe1 in position 8: invalid continuation byte
``` 
but when I changed it to 
```Python 
> b'CN=Rak V\xe1clav'.decode('windows-1250')
'CN=Rak Václav'
``` 

I made a simple patch, what solve my issue. I'll send you a merge request.

cannatag:
Hi Vaclav,
String in Ldap responses should always be in utf-8 as per the current RFCs. Which server are you connecting to?

Giovanni

vaclavrak:
I'm not server admin so I don't know this exactly. 

So far I know is MS Active Directory  LDAP  implementation on Microsoft Server.

cannatag:
Please check code in dev, I've slightly modified your patch. The config parameter is now called ``RESPONSE_DN_ENCODING`` and can be set to one encoding or a list of encodings. If a list is provided ldap3 tries sequentially until an ecoding doesn't return a UnicodeDecodeError or raise an LDAPDNError exception.

Can you try to search the offending entries and to get the cn attribute back? ldap3 still tries to decode the search response attributes with utf-8 (as per the standard RFC). If you can't get the cn back I will apply a similar parameter to attribute values too.

vaclavrak:
I like the way you update the patch. 

the other thing about CN filter. I did several test and I discovered two things
1. when I code string like:
```Python
>>> u'(CN=Rak Václav)'.encode('windows-1250')
b'(CN=Rak V\xe1clav)'
``` 
this was not suprise that result is *byte*, but this is not accepted by search function. Because python3 have natural encoding UTF-8 I could not use any kind of string 

2. if I put into search function CN parameter in UTF-8 string it doesn't find anyting. I also try to search some name without diacritics and it was sucessfull. 


during my test I found interesting thing in conv.py:43 is "DEFAULT_ENCODING" config parameter what solve it. If I set this into "windows-1250" it search correctly names with diacritics. 

[...]

PR #302, opened 2017-02-21
patch for MS implementations of LDAP 

vaclavrak:
here is small patch

1. change encoding for search results regarding  https://github.com/cannatag/ldap3/issues/301 
```Python
        set_config_parameter("SEARCH_RESULT_ENCODING", "windows-1250")
        with ldap3.Connection(server, user, password=pw, version=2, auto_bind=True) as conn:
            conn.search(basedn, u'(mail=user@domain.com)')
            print(conn.entries)
            print(server.info)
```

2. problem with reserved response from LDAP: this code is reserved and was missing in list. This invoke error on MS implementation of LDAP.


cannatag:
thanks. will work on this in the weekend.

cannatag:
HI,
I've checked your code. You're applying the SEARCH_RESULT_ENCODING only to the "dn" of the found entry and not to all the entries result. Are you sure this is what you want? If the server uses a different encoding than utf-8 (even if not permmited by the RFCs) it should apply it to all strings returned (that should be in utf-8 too).

Also I'm thinking of using a "cascading" list of encodings: for example if SEARCH_RESULT_ENCODING is set to {'utf-8', 'windows-1250'] it would try utf-8 first and if it get a decode error tries with 'windows-1250'.
This should be helpful if the server uses different encodings (I don't know if this could ever happen).

What do you think?

PR #306, opened 2017-03-01
Update modify.rst

pmarillo:
Remove erroneous parameter description ?


cannatag:
Thanks!

Will merge immediately.

Bye,
Giovanni

PR #317, opened 2017-03-14
Revert "Make ObjectDef use aliased attribute names, too"

cannatag:
Reverts cannatag/ldap3#316 - sorry the "attr" variable is not defined!


cannatag:
can you fix it?

PR #318, opened 2017-03-14
Corrected ciDict key

styopa:


cannatag:
it's working now. Thank you.

Giovanni

PR #320, opened 2017-03-15
Validate string representation of integers

styopa:
Integers could be formatted as strings, e.g. in simple queries.

Example query:
`'uidNumber: 42'`


cannatag:
I've modified your implementation. Please check code in dev. It should works now. I've also addes a test file for validators.

cannatag:
released in 2.2.2

PR #331, opened 2017-04-04
Fixed logic where checks would be made against schema.attribute_types…

glenbot:
… as None

In some cases, like `foxpass.com` ldap services the schema doesn't return any attribute types and there are `in` statements that are checking against `None` causing it to crash. Added guards around those areas.


cannatag:
thanks, will include this in the DEV brach. if test are ok I will include your changes in the next version of ldap3.

PR #349, opened 2017-05-11
Implements '!=' operator for attributes

fenrrir:
The missing ('__ne__') method produces unexpected results. The code below shows it:

`class T(object):

	def __init__(self, value):
		self.v = value

	def __eq__(self, other):
		return self.v == other

class T2(T):

	def __ne__(self, other):
		return not self == other

assert T(5) == 5
assert T(5) != 5

assert T2(5) == 5
assert (T2(5) != 5) == False`


cannatag:
Thanks, can you send the pull request to the "dev" branch? The master is for the latest production code and is changed only when a new version is released.

Bye,
Giovanni

PR #353, opened 2017-05-26
Aligned default configurable settings to documentation as the documented values are misleading.

guarnacciaa:
Even though the document doesn't explicitly report that the reported values are defaults, that's still very misleading as it leads to think that ldap3 will behave according to such values and it makes hard for unexperienced developers to understand why the time needed to handle LDAP responses is that "high" when the expectation set by the documentation is totally different .


cannatag:
Thanks, the values were probably changed in some tests.

Will fix it in the next version of ldap3. Can't merge because you sent the pull request to the master branch that reflects the last released version.  Can you send it to the dev branch?

cannatag:
Docs and parameters updated in dev

PR #379, opened 2017-07-04
Fixed error that resulted in valid generalizedTime strings not being parsed

Busuwe:
An error in format_time caused valid generalizedTime strings to fail to parse.
The valid strings that failed were strings with only the year, month, day and hour components, and no minute, second or fraction component.
>>> ldap3.protocol.formatters.formatters.format_time(b'2010093022Z')
b'2010093022Z'
Where the expected result should have been
datetime.datetime(2010, 9, 30, 22, 0, tzinfo=OffsetTzInfo(offset=0, name='UTC'))


cannatag:
Thanks!

Issue #385, opened 2017-07-12
LDAPSocketOpenError('socket ssl wrapping error: [SSL: WRONG_VERSION_NUMBER] wrong version number 


katul27:
I am getting above error for below tls ctx,
tls = Tls(local_private_key_file = 'www.example.com.key.pem',
              local_certificate_file = 'www.example.com.cert.pem',
              validate = ssl.CERT_OPTIONAL,
              version = ssl.PROTOCOL_TLSv1_2,
               ca_certs_file = 'ca-chain.cert.pem')

 have below version of
openssl: OpenSSL 1.0.1e-fips
python: Python 2.7.5

i can understand that this error is coming because of some version mismatch, can some guide me to resolving this issue.

cannatag:
Are you sure that the server you're connecting to is listening in SSL mode?
This error commonly occurs when trying to connect with SSL on a socket that is not negotiating SSL.

katul27:
My local LDAP test server will wait on TCP connection, once it will receive extended TLS request, then it will go to TLS handshake, after that i m doing connection.start_tls(), but again getting same error,

thinking some sequence is missing. Can guide me the creating TLS connection with LDAP server.

cannatag:
What happens if you try without any Tls() configuration? ldap3 has a Tls default configuration that let you accept any certificate. Please try the following:

```
from ldap3 import Server, Connection
s = Server('my_server')
c = Connection(s, 'my_user', 'my_password')
c.bind()
print(c)  # to check the connection status
c.start_tls()
print(c)  # to check the connection status
...
```

send me the output of this snippet of code with server, username and password adjusted to your configuration.

Bye,
Giovanni

katul27:
I am running below script,

#!/usr/bin/python

# import class and constants
import ssl
from ldap3 import Server, Connection, Tls, ALL

# define the server
t = Tls(validate=ssl.CERT_NONE)
server = Server(host='localhost', port=*********, use_ssl=True, tls=t)
# define the connection
c = Connection(server, user='****************', password='*************', auto_bind='NONE', version=3, authentication='SIMPLE')
c.start_tls()
c.bind()
print(c.result)
#close connection
c.unbind()

Getting below error, but I run python LDAP module same thing then it is working, so I am not able to figure out what and where is the problem.

katul27:
Output of above script 
[...]

PR #387, opened 2017-07-17
Ensure optional ASN.1 values presence

etingof:
Following up [this issue](https://github.com/cannatag/ldap3/issues/385), here is a quick fix aiming at making sure the OPTIONAL ASN.1 value is indeed supplied by the remote end prior to using the value.

Grepping through the sources, I can see more `OptionalNamedType` fields being used. I guess we'd need to apply similar condition to each "read" from these optional fields. WDYT?

### DIFF
--- a/ldap3/operation/extended.py
+++ b/ldap3/operation/extended.py
@@ -55,7 +55,7 @@ def extended_operation(request_name,
 
 
 def extended_request_to_dict(request):
-    return {'name': str(request['requestName']), 'value': bytes(request['requestValue']) if request['requestValue'].hasValue() and request['requestValue'] else None}
+    return {'name': str(request['requestName']), 'value': bytes(request['requestValue']) if request['requestValue'] is not None and request['requestValue'].hasValue() else None}
 
 
 def extended_response_to_dict(response):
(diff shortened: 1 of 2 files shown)


cannatag:
Hi Ylya,
with your code I get the following error (for extended request):

```
  File "C:\Python\OpenSource\ldap3\ldap3\operation\extended.py", line 61, in extended_request_to_dict
    return {'name': str(request['requestName']), 'value': bytes(request['requestValue']) if request['requestValue'].hasValue() and request['requestValue'] else None}
AttributeError: 'NoneType' object has no attribute 'hasValue'
```
because request['requestValue'] is simply None instead of a pyasn1 value. Is this correct? In this case I have to check before for request['requestValue'] existance and then for having a value:

```
return {'name': str(request['requestName']), 'value': bytes(request['requestValue']) if 'requestValue' in request and request['requestValue'] and request['requestValue'].hasValue()  else None}
```

etingof:
Hi Giovanni!

You are right, I've been targeting the latest (yet unreleased) pyasn1 which always returns a ASN.1 type or value for optional fields, not `None` as it is with the older pyasn1.

I've fixed my patch to handle both cases.

cannatag:
version 2.3 of ldap3 is now compatible with pyasn1 from version 0.1.8 up to 0.3.1.

PR #392, opened 2017-08-02
Check for password correctly in rebind

dmanchon:
Problem: Rebind is comparing making this comparison to validate input data:
```python
if password:
    self.password = password
```
That condition is false in case of `password = ''`, so the old value of the object will be keep in that case, including case where the bind was correct.

This is a potential security hole, since a correct bind followed by a rebind with empty password will complete the binding correctly.

I have changed the way the password is checked to only check if None.

### DIFF
--- a/ldap3/core/connection.py
+++ b/ldap3/core/connection.py
@@ -607,7 +607,7 @@ class Connection(object):
         with self.lock:
             if user:
                 self.user = user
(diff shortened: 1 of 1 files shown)


cannatag:
Please don't send PR to the master branch. I will add your PR manually to the dev branch.

PR #401, opened 2017-08-23
Sasl kerberos py3 compatibility

Busuwe:
When authenticating using SASL/Kerberos in Python3 a UnicodeDecodeError is raised.

The culprit seems to be the str conversion done on sasl['credentials'] in the sasl_to_dict function in ldap3.protocol.convert.py, and should instead be a bytes conversion as that will return str in python 2 and bytes in 3. 
With that in place binding using SASL/Kerberos works in both Python 2 and 3.

Software versions tested:
Python 2.7.6, 3.4.3 and 3.6.0
gssapi 1.2.0
ldap3  2.3.0
pyasn1 0.3.2

Example code:

```
server = ldap3.Server("ldap.example.com", use_ssl=True, port = 636)
conn = ldap3.Connection(server, authentication=ldap3.SASL, sasl_mechanism=ldap3.KERBEROS, raise_exceptions=True)
conn.bind()

  File "/usr/local/lib/python3.4/dist-packages/pyasn1/type/univ.py", line 978, in __str__
    return self._value.decode(self.encoding)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x82 in position 1: invalid start byte

During handling of the above exception, another exception occurred:

  File "mini.py", line 15, in <module>
    conn.bind()
  File "/mnt/c/Users/mattias.hedlund/git/ldap3/ldap3/core/connection.py", line 538, in bind
    response = self.do_sasl_bind(controls)
  File "/mnt/c/Users/mattias.hedlund/git/ldap3/ldap3/core/connection.py", line 1233, in do_sasl_bind
    result = sasl_gssapi(self, controls)
  File "/mnt/c/Users/mattias.hedlund/git/ldap3/ldap3/protocol/sasl/kerberos.py", line 84, in sasl_gssapi
    result = send_sasl_negotiation(connection, controls, out_token)
  File "/mnt/c/Users/mattias.hedlund/git/ldap3/ldap3/protocol/sasl/sasl.py", line 160, in send_sasl_negotiation
    response = connection.post_send_single_response(connection.send('bindRequest', request, controls))
  File "/mnt/c/Users/mattias.hedlund/git/ldap3/ldap3/strategy/base.py", line 300, in send
    self.connection.request = BaseStrategy.decode_request(message_type, request, controls)
  File "/mnt/c/Users/mattias.hedlund/git/ldap3/ldap3/strategy/base.py", line 615, in decode_request
    result = bind_request_to_dict(component)
  File "/mnt/c/Users/mattias.hedlund/git/ldap3/ldap3/operation/bind.py", line 96, in bind_request_to_dict
    'authentication': authentication_choice_to_dict(request['authentication'])}
  File "/mnt/c/Users/mattias.hedlund/git/ldap3/ldap3/protocol/convert.py", line 62, in authentication_choice_to_dict
    return {'simple': str(authentication_choice['simple']) if authentication_choice.getName() == 'simple' else None, 'sasl': sasl_to_dict(authentication_choice['sasl']) if authentication_choice.getName() == 'sasl' else None}
  File "/mnt/c/Users/mattias.hedlund/git/ldap3/ldap3/protocol/convert.py", line 58, in sasl_to_dict
    return {'mechanism': str(sasl['mechanism']), 'credentials': str(sasl['credentials'] if sasl['credentials'] is not None and sasl['credentials'].hasValue() else None)}
  File "/usr/local/lib/python3.4/dist-packages/pyasn1/type/univ.py", line 982, in __str__
    'Can\'t decode string \'%s\' with \'%s\' codec at \'%s\'' % (self._value, self.encoding, self.__class__.__name__)
pyasn1.error.PyAsn1Error: Can't decode string 'b'`\x82\x02\x8b\x06<omitted rest of the blob>' with 'utf-8' codec at 'Credentials'
```


etingof:
You may want to use `sasl['credentials'].asOctets()` which would automatically call `str()` on Py2 and `bytes()` on Py3.

More importantly, is this `credentials` field guaranteed to always contain the UTF-8 serialisation of the unicode characters?

If yes, the problem seems to be on the server side which somehow serialzied the unicode string badly. So we should probably catch utf-8 deserialisation failures gracefully.

If those credentials are not unicode, then [we should not pretend they are](https://github.com/cannatag/ldap3/blob/master/ldap3/protocol/rfc4511.py#L179).

Busuwe:
The credentials field can contain arbitrary data depending on the SASL mechanism used, so not just valid utf8.

DIGEST, PLAIN and GSSAPI/KERBEROS all prepare byte-like objects which are then assigned to sasl['credentials']. DIGEST and PLAIN both produce something that can be decoded back into utf8. GSSAPI/KERBEROS doesn't and thus raises UnicodeDecodeError with the str conversion. Stripping the str conversion entirely also seems to solve the problem as sasl['credentials'] is already a subclass of OctetString. 

EXTERNAL is however assigning the user supplied str directly to to sasl['credentials']. If no encoding is specified in the Credentials class, it will assume that the (unicode) string should be decoded using latin-1 which would be incorrect. 
One could of course argue that EXTERNAL should explicitly encode it before passing it to send_sasl_negotiation() in ldap3.protocol.sasl.external.sasl_external(), thus removing the need for the implicit conversion in the assignment to sasl['credentials'], and by extension the need for the encoding = "utf8" on Credentials.

etingof:
> If no encoding is specified in the Credentials class, it will assume that the (unicode) string should 
> be decoded using latin-1 which would be incorrect.

My understanding is that the [iso-8859-1](https://github.com/etingof/pyasn1/blob/master/pyasn1/type/univ.py#L806) encoding is a no-op. It will never interpret/modify the data being transformed, so it will just change type.

That is why, I guess, you can safely turn chars to ordinals (bytes) and back by en/decoding your data through the `iso-8859-1` codepage.

cannatag:
I merge this PR, can you check code in dev to be sure it's working as requested? I don't use Kerberos authentication at all, sorry.

cannatag:
@etingof @Busuwe What should we do with the Credential encoding? I agree with Ilya that the encoding should be removed because it's not specified in the ASN.1 definition.

Busuwe:
I agree.
I believe we just need to take care to encode connection.sasl_credentials when it's passed to send_sasl_negotiation() in the function ldap3.protocol.sasl.external.sasl_external().
The other mechanisms already do so.

PR #402, opened 2017-08-23
Sasl kerberos authzid

Busuwe:
This change allows a user to request a different authorization ID when performing a bind using Kerberos.
The authorization ID is passed as the second element in the sasl_credentials tuple.
No change in the behavior from previously if sasl_credentials is a 1-tuple or None

Tested in Python 2.7.6 and 3.4.3. (py3 requires the fix in #401)


cannatag:
Thanks!

Issue #415, opened 2017-09-25
who_am_i() not working with the mocked strategies


Busuwe:
When using the connection.extended.standard.who_am_i() function, an error is thrown when using the MOCK_SYNC strategy.

```
import ldap3

# Basic info and schema from a relatively plain OpenLDAP server,
# as who_am_i requires "1.3.6.1.4.1.4203.1.11.3" in supportedExtension
mocked_server = ldap3.Server.from_definition("fake_server", "testinfo.json", "testschema.json")

conn = ldap3.Connection(mocked_server, user='cn=test_user,dc=domain,dc=local',
                            password="secret_password", client_strategy=ldap3.MOCK_SYNC)

mocked_entries = {
    "dc=domain,dc=local": {"dc": ["domain"], "objectClass": ["dcObject"]},
    "cn=test_user,dc=domain,dc=local": {
        "cn": ["test_user"], "userPassword": ["secret_password"],
        "objectClass": ["user"],
    }
}

for key in mocked_entries:
    value = mocked_entries[key]
    conn.strategy.add_entry(key, value)

conn.bind()
bound_authz_dn = conn.extend.standard.who_am_i()
Traceback (most recent call last):
  File "mini.py", line 24, in <module>
    bound_authz_dn = conn.extend.standard.who_am_i()
  File "C:\Users\mattias.hedlund\PycharmProjects\klarnaLDAPv3\ldap3\extend\__init__.py", line 65, in who_am_i
    controls).send()
  File "C:\Users\mattias.hedlund\PycharmProjects\klarnaLDAPv3\ldap3\extend\operation.py", line 53, in send
    resp = self.connection.extended(self.request_name, self.request_value, self.controls)
  File "C:\Users\mattias.hedlund\PycharmProjects\klarnaLDAPv3\ldap3\core\connection.py", line 1168, in extended
    response = self.post_send_single_response(self.send('extendedReq', request, controls))
  File "C:\Users\mattias.hedlund\PycharmProjects\klarnaLDAPv3\ldap3\strategy\mockSync.py", line 98, in post_send_single_response
    result = extended_response_to_dict(self.mock_extended(request, controls))
  File "C:\Users\mattias.hedlund\PycharmProjects\klarnaLDAPv3\ldap3\operation\extended.py", line 70, in extended_response_to_dict
    'responseValue': bytes(response['responseValue']) if response['responseValue'] is not None and response['responseValue'].hasValue() else bytes()}
[...]

PR #416, opened 2017-09-25
Mocked who am i

Busuwe:
Potential fix for #415

The function [extended_response_to_dict](https://github.com/cannatag/ldap3/blob/dev/ldap3/strategy/mockSync.py#L98) seems to expect response['responseValue'] to be a OctetString, while mock_extended(...) is returning a BER encoded OctetString.


thatarchguy:
Confirmed working for me. I was running into the same errors as you in https://github.com/cannatag/ldap3/issues/415. Thank you!
```
pip install https://github.com/Busuwe/ldap3/archive/a94fb4.zip --upgrade
```

PR #421, opened 2017-10-16
Provide a `connected` attribute to `Connection`

s0undt3ch:


cannatag:
Hi, are you sure you need this? It is always the opposite of the connection.closed property. It also clashes with the connection.bound property that is true when the connection is open AND a successful bind has been performed.

s0undt3ch:
Yeah, let's close.

PR #422, opened 2017-10-16
Allow installing ldap3 compiled to C using cython

s0undt3ch:
There's no cython specific code in the package so we don't take much
advantage from cython, but running from compiled C code, even in this
scenario has proven to be at least a bit more performant.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@ __pycache__/
 
 # C extensions
 *.so
+# Cythonized .c files
+*.c
 
 # Distribution / packaging
 .Python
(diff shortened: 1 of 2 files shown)


cannatag:
Thanks for the patch. I don’t use Cython so I can’t test it.

s0undt3ch:
I noticed a Travis file but didn't see this running in Travis.
Best scenario, the existing matrix would run twice, non compiled and then compiled...

cannatag:
The Travis-CI link is https://travis-ci.org/cannatag/ldap3/builds/290445414?utm_source=email&utm_medium=notification

PR #423, opened 2017-10-19
Fixed bug that would evaluate filters with an AND incorrectly

signedbit:
Fixed bug that would evaluate filters with an AND incorrectly when the first operand evaluates to false.

For example, the following filters would return different results: 
* `(&(userPassword=notarealvalue)(revision=1))`
* `(&(revision=1)(userPassword=notarealvalue))`


cannatag:
Thanks! Will be released in 2.4.

PR #425, opened 2017-10-30
Code fix: sending an invalid cookie is an error

moonpatrol:
Sending "new_cookie" as paged_cookie on an initial request is an error when talking to a 389ds ldap server. My code only sends a paged_cookie on subsequent requests.


cannatag:
thanks!

PR #450, opened 2017-12-12
Pass sasl_* variables to referral connection

BlackVoid:
This PR passes the sasl_* args added in #402 to the referral connection.

Tested in Python 3.4.5


cannatag:
Thanks!

Issue #100, opened 2015-08-14
Samba does not honor delete_old_dn=True


w-p:
When attempting to rename an object with __connection.modify_dn()__, an error is raised with the message __ERROR, 00002035: Unwilling to perform. Old RDN must be deleted__ on Samba domain controllers.  With full debug logging on, it appears that the correct information is being transmitted to the server. This operation has been tested against Microsoft Server 2008 R2 and OpenLDAP with success.

I've had similar issues in the past with the python-ldap library but currently I cannot find the code to illustrate resolution.  This issue may be specifically Samba related but thought it should be noted here.

w-p:
Possibly related?
http://stackoverflow.com/questions/21033254/ruby-netldap-move-user-entry-from-one-container-to-another-in-samba4

cannatag:
Can you try with delete_old_dn=False? The rfc is quite vague on the meaning of this field. it's not clear what the meaning of 0 and 1 (the valued encoded in the message) is.

Bye,
Giovanni

w-p:
I've attempted:
> bool: True, False
> int: 1, 0
> byte: b'1', b'0'
> hex: 0x1, 0x0

All result in:
> diagnosticMessage=b'00002035: Unwilling to perform. Old RDN must be deleted'

Again, the operation works for without issue against Active Directory, just not Samba.  On the other hand, the python-ldap library executes a rename with similar arguments without issue.  I spent a little time looking through the Samba, python-ldap, and OpenLDAP code bases but saw nothing obvious.

I'll retest later today with another MS server and try an OpenLDAP as well.

w-p:
Just tested on my OpenLDAP server with success.

cannatag:
Will, are you able to send me a trace of the succesful operation to openLDAP and a trace to the unsuccesful connection to Samba? You can get them with tcpdump or wireshark. The connection must be not encrypted.

If you're not able to do so I will try to set up a lab with samba, but it will take a few days for me to do that.

Bye,
Giovanni

w-p:
I will try to do that today.  I also opened an issue with the Samba guys.
https://bugzilla.samba.org/show_bug.cgi?id=11453

cannatag:
can you tell me your samba environment? how are you connecting to the samba server? are you using an ldap server for authentication? I would try to recreate your environement. thanks

w-p:
[...]

PR #463, opened 2018-01-03
 boolean asn1 monkeypatch update

tmarlok88:
this fix: https://github.com/cannatag/ldap3/issues/100 does not work anymore because the pyasn1 library has changed since. I made some adjustments to make it work with the current pyasn1 library.

### DIFF
--- a/ldap3/utils/asn1.py
+++ b/ldap3/utils/asn1.py
@@ -1,6 +1,3 @@
-typeMap,"""
-"""
-
 # Created on 2015.08.19
 #
 # Author: Giovanni Cannata


etingof:
Note that this patch [may not really work](https://github.com/etingof/pyasn1/blob/v0.3.7/pyasn1/codec/cer/encoder.py#L16) with the reasonably fresh pyasn1 versions.

So may be you should monkey-patch pyasn1 conditionally (based on the actual `pyasn1.__version__`) or just bump ldap3 requirements to pyasn1>=0.3.7?

tmarlok88:
The file you linked is a cer encoder whereas the monkey-patched file is a ber encoder. And as far as I see that hasn't changed since the 0.3.1.

etingof:
Oh, you are right, that's my bad!

It looks like, with BER, it is allowed to use any non-zero value for `True` encoding. May be we should just move pyasn1's  `Boolean` encoder from CER into DER?

Any concerns? ;-)

tmarlok88:
Well, I'm not really familiar with the code, but it looks like a fair solution to me.

cannatag:
Hi @tmarlok88  and @etingof ,
the problem is that  the LDAP protocol uses the BER encoding except for the Boolean value that must be 00 for False and FF for True. So I prefer to not globally switch to CER or DER encodings, that are more restrictive than BER, having only one possible option for a value, because this could work when sending data but not when receiving. The old monkeypatch was needed to restrict the boolean value to FF only when sending, while it allowed any non 0 value when receiving.

etingof:
> the problem is that the LDAP protocol uses the BER encoding except for the Boolean value that must be 00 for False and FF for True. 

Is it enforced in some LDAP RFCs or is it a de-facto situation?

> So I prefer to not globally switch to CER or DER encodings, that are more restrictive than BER, having only one possible option for a value, because this could work when sending data but not when receiving. 

What I meant is to change pyasn1 **encoder** to encode True as FF rather than as 01. But decoder should still treat any non-zero value as True.

Note that BER standard allows any non-zero octet to encode Boolean True value (01 is currently used in pyasn1) while CER specifically requires FF.

cannatag:
@etingof 
This is, unfortunately, stated in the official LDAP RFC [RFC4511](https://www.ietf.org/rfc/rfc4511.txt): 

> 
> 5.1.  Protocol Encoding
> 
>    The protocol elements of LDAP SHALL be encoded for exchange using the
>    Basic Encoding Rules [BER] of [ASN.1] with the following
>    restrictions:
> 
>    - Only the definite form of length encoding is used.
> 
>    - OCTET STRING values are encoded in the primitive form only.
> 
>    - If the value of a BOOLEAN type is true, the encoding of the value

etingof:
So weird! Why they choose to constrain the lower levels of the stack which they have no control of...?

Then I can only offer adding a  class attribute to pyasn1 *BooleanEncoder* holding the constant to be used for encoding. Something along the lines of:

```
class BooleanEncoder(AbstractItemEncoder):
    supportIndefLenMode = False
    truthValue = 0x01

    def encodeValue(self, value, asn1Spec, encodeFun, **options):
        return value and (self.truthValue,) or (0,), False, False
```

That may slightly simplify your patch at the cost of stricter pyasn1 version requirement in the dependencies. Let me know if you want that. ;-)

etingof:
Other minor note is that instead of monkey-patching pyasn1 you could copy its `typeMap` and `tagMap`, then pass them to `encode.encoder()` whenever ldap3 calls it.

cannatag:
HI @etingof, can you check the following code? It's almost identical to the original patch posted by @tmarlok88:

```
from pyasn1.codec.ber.encoder import AbstractItemEncoder, encode, tagMap, typeMap
from pyasn1.type import univ

class LDAPBooleanEncoder(AbstractItemEncoder):
    supportIndefLenMode = False

    def encodeValue(self, value, asn1Spec, encodeFun, **options):
        substrate = (0,) if value == 0 else (255,)
        return substrate, False, False

tagMap[univ.Boolean.tagSet] = LDAPBooleanEncoder()
typeMap[univ.Boolean.typeId] = LDAPBooleanEncoder()

print(encode(univ.Boolean(True)))
```
Or do you think that a new Encoder() should be created explicitly with modified copy of tagMap and typeMap:

```
from pyasn1.codec.ber.encoder import AbstractItemEncoder, Encoder, tagMap, typeMap
from pyasn1.type import univ
from copy import deepcopy

class LDAPBooleanEncoder(AbstractItemEncoder):
    supportIndefLenMode = False

    def encodeValue(self, value, asn1Spec, encodeFun, **options):
        substrate = (0,) if value == 0 else (255,)
        return substrate, False, False

customTagMap = deepcopy(tagMap)
customTypeMap = deepcopy(typeMap)
customTagMap[univ.Boolean.tagSet] = LDAPBooleanEncoder()
customTypeMap[univ.Boolean.typeId] = LDAPBooleanEncoder()

encode = Encoder(customTagMap, customTypeMap)
print(encode(univ.Boolean(True)))
```

The second solution don't change the original pyasn1 behaviour so probably could be used with older and newer version of pyasn1. 

@etingof  What do you think?

tmarlok88:
Hi @cannatag!
I think the second version might be a better choice. I will update my PR if you agree.

cannatag:
I have modified the code with the second solution, it seem better to me.

Can you check the code in dev?

Thanks,
Giovanni

cannatag:
Please be aware that the current patch doesn't work with pyasn1 <= 0.3.7. I'm working on this.

tmarlok88:
i am currently away from keyboard for a week. will test it after that.

On Jan 14, 2018 7:27 PM, "Giovanni Cannata" <notifications@github.com>
wrote:

Please be aware that the current patch doesn't work with pyasn1 <= 0.3.7.
I'm working on this.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/cannatag/ldap3/pull/463#issuecomment-357531191>, or mute
the thread

cannatag:
I've adjust the code for pyasn1 from 0.1.8 to 0.4.2, there were 3 variants. It is in dev now.

cannatag:
@etingof please can you check if the code is correct? I've found that there are 5 different signatures for encodeValue to care of in different versions of pyasn1:

```
        if pyasn1_version <= '0.2.3':
            from pyasn1.compat.octets import ints2octs
            _true = ints2octs((255,))
            _false = ints2octs((0,))
            def encodeValue(self, encodeFun, value, defMode, maxChunkSize):
                return value and self._true or self._false, 0
        elif pyasn1_version <= '0.3.1':
            def encodeValue(self, encodeFun, value, defMode, maxChunkSize):
                return value and (255,) or (0,), False, False
        elif pyasn1_version <= '0.3.4':
            def encodeValue(self, encodeFun, value, defMode, maxChunkSize, ifNotEmpty=False):
                return value and (255,) or (0,), False, False
        elif pyasn1_version <= '0.3.7':
            def encodeValue(self, value, encodeFun, **options):
                return value and (255,) or (0,), False, False
        else:
            def encodeValue(self, value, asn1Spec, encodeFun, **options):
                return value and (255,) or (0,), False, False
```

@etingof  Does this seem right to you?

Thanks, 
Giovanni

etingof:
I think we can reduce these cases to just two:

        if pyasn1_version <= '0.2.3':
            from pyasn1.compat.octets import ints2octs
            _true = ints2octs((255,))
            _false = ints2octs((0,))
            def encodeValue(self, encodeFun, value, defMode, maxChunkSize):
                return value and self._true or self._false, 0
        else:
            def encodeValue(self, encodeFun, value, defMode, maxChunkSize, **options):
                return value and (255,) or (0,), False, False

Does it work for you?

I am sorry for it is getting messy - the thing is that we are dealing with internal pyasn1 API, it is not supposed to be stable. ;-)

cannatag:
Hi @etingof, thanks for your suggestion. Are you sure this will work with version from 0.3.5 to 0.3.7? The value and encodeFun parameter are inverted. I've tested it and get no error but I'm not sure if ANDing an object and a tuple can raise errors with some future python version.

cannatag:
@etingof With your code and pyasn 0.3.6 I get the following error:
```
0.3.6
  File "C:/Users/Giovanni Cannata/.PyCharm2017.3/config/scratches/scratch_22.py", line 23, in <module>
    print(list(encode(Boolean(True))))
  File "C:\Python\Python36\lib\site-packages\pyasn1\codec\ber\encoder.py", line 516, in __call__
    substrate = concreteEncoder.encode(value, self, **options)
  File "C:\Python\Python36\lib\site-packages\pyasn1\codec\ber\encoder.py", line 77, in encode
    value, encodeFun, **options
TypeError: encodeValue() missing 2 required positional arguments: 'defMode' and 'maxChunkSize'

Process finished with exit code 1
```

I think it's better to be more explicit and to catch all the possibile variations of the signature...

etingof:
Oh, well. Your original code should work just fine. I've been trying to shorten it. ;-/

cannatag:
Released in version 2.4.1

PR #475, opened 2018-01-17
Add parentheses failing test for parse_filter

akosiaris:
The state machine in parse_search_filter seems to not handle the
situation where the filter may have for any reason parenteses not
related to grouping terms. Add a test for that to showcase the issue

### DIFF
--- a/test/testParseSearchFilter.py
+++ b/test/testParseSearchFilter.py
@@ -26,6 +26,7 @@
 import unittest
 
 from ldap3.operation.search import parse_filter, MATCH_EQUAL, MATCH_EXTENSIBLE
+from ldap3.utils.conv import escape_filter_chars
 from test.config import test_auto_escape, test_auto_encode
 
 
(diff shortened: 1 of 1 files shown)


akosiaris:
Just run into this, which I consider a bug (but do correct me if I am wrong), so I 've created a test case to showcase it.

cannatag:
Hi, there are a few characters (parentheses, backslash, asterisk and the null char) that must be escaped in LDAP filters. You can use the ``ldap3.conv.escape_filter_chars()`` function to properly escape them:

```
from ldap3.utils.conv import escape_filter_chars
    def test_parse_search_filter_parenteses(self):
        f = parse_filter('(cn=' + escape_filter_chars('Doe (Missing Inc)') + ')', None, test_auto_escape, test_auto_encode)
        self.assertEqual(f.elements[0].tag, MATCH_EQUAL)
        self.assertEqual(f.elements[0].assertion['attr'], 'cn')
        self.assertEqual(f.elements[0].assertion['value'], b'Doe \\28Missing Inc\\29')
```

Now the code works as expected. Anyway I will include this thes in the test suite. 

Thanks.
Giovanni

akosiaris:
Ah, thanks I hadn't seen that function.

I have to admit I would expect from `parse_filter()` (actually I expect it from `search()` but anyway) to do the `escape_filter_chars()` dance for me and not have to do it manually. From your example above, I am guessing the reason this doesn't happen is the difficulty in figuring out which part is actual term grouping parenthesis and which are "data".

I 've amended the PR to look like your example as you mentioned wanting to include it.

cannatag:
Yes, it should be very hard to parse the filter with correct parentheses. There is an auto_escape feature in ldap3 but doesn't fix the parentheses. Sorry.

cannatag:
Thanks for the test code. I've included it in the dev branch.

PR #481, opened 2018-01-24
validate SSL by default

flokli:
It should be good practice in 2018 to validate SSL certificates by default, and require the user to disable this explicitly if a validation should NOT be performed.

### DIFF
--- a/ldap3/core/tls.py
+++ b/ldap3/core/tls.py
@@ -174,7 +174,7 @@ class Tls(object):
                                                      cafile=self.ca_certs_file,
                                                      capath=self.ca_certs_path,
                                                      cadata=self.ca_certs_data)
-            else:  # code from create_default_context in the Python standard library 3.5.1, creates a ssl context with the specificd protocol version
+            else:  # code from create_default_context in the Python standard library 3.5.1, creates a ssl context with the specified protocol version
                 ssl_context = ssl.SSLContext(self.version)
                 if self.ca_certs_file or self.ca_certs_path or self.ca_certs_data:
                     ssl_context.load_verify_locations(self.ca_certs_file, self.ca_certs_path, self.ca_certs_data)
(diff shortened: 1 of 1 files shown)


cannatag:
Hi, sorry for answering late. I'm not sure if this is a feasible approach. LDAP very often is used in internal network where certificates are self signed. To check for ssl validation would cause troubles to a lot of people. Perhaps is better to include a clear statement of this in the docs.

cannatag:
I prefer not to change this default because it breaks compatibility with previous v2.x version because I'm following semantic versioning rules. I can change in version 3.x.

flokli:
@cannatag understood. Can we still add an issue to a 3.x milestone, so this won't get lost?

cannatag:
Sure. I’ll create a version 3 milestone and add this as a feature.

Thanks,
Giovanni

Augustin-FL:
I'm sorry to insist but I definitely agree with this PR : would it be possible to enforce certificate checking by default?

If users are using this lib in an internal network, then it's the job of these users to set up an internal PKI and to trust it (in the Windows Certificate Trust Store, or in the Linux folders `/etc/ssl/certs` and `/etc/pki/ca-trust/ `)

cannatag:
I know, but this would break a lot of running code. For example most of my programs that use the ldap3 library run in Intranets where ldap servers are using self signed certificates and they would stop working if a newer version of the library validates the certificate by default.

This would be a breaking change that require a major release, so I think it will be in version 3 of ldap3.

I will open an issue for suggestion on version 3 new features and depreciations (probably version 3 will abandon the python 2.6 and python 2.7 compatibility).

lhoekenga:
I understand that this is a breaking change, but this is irresponsible.

The default should be certificate validation with a mechanism to turn off the validation requirements, i.e. openldap's `TLS_REQCERT = never`

where's version 3?

PR #489, opened 2018-02-02
Wrong error raised for SIMPLE auth when user is empty

vandot:
When user is empty LDAPPasswordIsMandatoryError is raised and it should be LDAPUserNameIsMandatoryError


cannatag:
thanks!

PR #491, opened 2018-02-06
Make entry membership check return false if entry doesn't contain item key

maxsond:
The EntryBase class's __contains__ method checks if an entry contains an item by trying to get it with __getitem__. If the attribute is not found by line 234, then __getitem__ raises an LDAPCursorError exception with the message that the key was not found.

It seems that if the item's key was not found, then __contains__ should return False rather than propagating the exception.

### DIFF
--- a/ldap3/abstract/entry.py
+++ b/ldap3/abstract/entry.py
@@ -147,7 +147,7 @@ class EntryBase(object):
         try:
             self.__getitem__(item)
             return True
-        except LDAPKeyError:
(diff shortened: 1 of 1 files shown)


maxsond:
It might be that __getitem__ should instead raise a KeyError exception if it fails to find the key, in which case the change in this pull request is unnecessary. Looking back at branch 2.2.3, the __getitem__ method used to raise a KeyError when it failed to find the key, but this has since been changed to an LDAPCursorError.

PR #497, opened 2018-02-13
Add handling of size limit exceed on mocked connection

dkaatz:
The `mockBase` is actually not handling the size limit exceeded in case `raise_exceptions=True` on the connection object.

I've added the necessary handling and a test for this.

Let me know if you want me to change anything.

Kind Regards.


cannatag:
Thanks for your precious work!

PR #508, opened 2018-03-14
Fix paged_search_generator in combination with size_limit

fredreichbier:
Hi Giovanni,

since ldap3 2.3, ``paged_search_generator`` raises a ``LDAPSizeLimitExceededResult`` when the client-side size limit has been reached. While this wasn't the case before 2.3, I think this is fine behavior from an API design perspective. However, they seem to be situations in which the generator "loses" some entries, as illustrated by the following script:

```python
import ldap3
from ldap3.core.exceptions import LDAPSizeLimitExceededResult

SERVER = 'ipa.demo1.freeipa.org'
BIND_DN = 'uid=admin,cn=users,cn=accounts,dc=demo1,dc=freeipa,dc=org'
BIND_PW = 'Secret123'
BASE_DN = 'dc=demo1,dc=freeipa,dc=org'
FILTER = '(objectClass=person)'

server = ldap3.Server(SERVER, get_info=ldap3.ALL)
conn = ldap3.Connection(server, BIND_DN, BIND_PW, auto_bind=True)

entries = conn.extend.standard.paged_search(BASE_DN,
                                            FILTER,
                                            size_limit=3)

# Cannot use ``for entry in entries:`` because ``LDAPSizeLimitExceededResult`` is raised
length = 0
while True:
    try:
        entry = next(entries)
        length += 1
    except StopIteration:
        break
    except LDAPSizeLimitExceededResult:
        break
print '=> {} entries retrieved from generator'.format(length)
print '({} entries in conn.response)'.format(len(conn.response) if conn.response else 0)
```

As we set ``size_limit=3``, I would expect the script to retrieve three entries from the genereator. However, it prints
```
=> 0 entries retrieved from generator
(3 entries in conn.response)
```

I think this is because the exception is raised "too early" in ``paged_search_generator``, i.e. before the three entries waiting in ``conn.response`` have been processed.

I have attached a patch which fixes this by returning all entries from ``conn.response`` before raising the exception. However, I'm not sure if that's the best fix. What do you think?

Thanks and Best Regards

Friedrich

### DIFF
--- a/ldap3/extend/standard/PagedSearch.py
+++ b/ldap3/extend/standard/PagedSearch.py
@@ -25,7 +25,7 @@
 
 from ... import SUBTREE, DEREF_ALWAYS
 from ...utils.dn import safe_dn
-from ...core.results import DO_NOT_RAISE_EXCEPTIONS
+from ...core.results import DO_NOT_RAISE_EXCEPTIONS, RESULT_SIZE_LIMIT_EXCEEDED
 from ...core.exceptions import LDAPOperationResult
 from ...utils.log import log, log_enabled, ERROR, BASIC, PROTOCOL, NETWORK, EXTENDED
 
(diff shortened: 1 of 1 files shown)


cannatag:
It looks good to me, thanks!

PR #519, opened 2018-04-09
Fix duplicate controls

dirkjanm:
The current version has a bug that if `controls` are specified to a paged search, the page control is appended to the controls. This works only for the first page since the second page will have 2 `paged_search_control`s in the `controls` list and results in an `LDAPProtocolErrorResult`: `Error processing control`.
The patch provided copies the list to a new object and thus the `.append` will not modify it, and the original `controls` object will be used again for the next page.

### DIFF
--- a/ldap3/core/connection.py
+++ b/ldap3/core/connection.py
@@ -747,6 +747,9 @@ class Connection(object):
 
                 if controls is None:
                     controls = []
+                else:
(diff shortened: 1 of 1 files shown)


cannatag:
Thanks! 
Will include it in the upcoming 2.5 version of ldap3.

PR #526, opened 2018-04-21
Doc updates for how to get dn

fedorareis:
The current information for getting a dn is not correct, instead of `entry._dn` as the documentation currently says it should be `entry.entry_dn`


cannatag:
Thank you!

PR #540, opened 2018-06-08
pass exceptions from PooledConnectionThread to client thread

kxt:
Exceptions thrown during open(), start_tls() and bind() can reach the
top of the call stack of the pool thread without the client thread being
notified. This results in a traceback and makes handling those errors by
the client impossible.

This patch catches those exceptions too, and passes them to the client
thread just like errors arising while performing LDAP requests.


cannatag:
Thanks, it will be released in ldap3 v. 2.5.1.

PR #550, opened 2018-07-09
Add more detailed info about invalid value when using Mock strategy

jannero:
The error message "added values must be bytes if no offline schema is
provided in Mock strategies" is not very informative when you get
that error while writing tests for your own code; you need to dive
into ldap3 implementation to see that the real issues is that the
type of some field has invalid type (not numeric, string or sequence
type).


cannatag:
Thanks! Will be released in ldap3 v.2.5.1

PR #551, opened 2018-07-11
malformatted GUID when used in filters

sonicsqwirl:
added condition to take into consideration if 'x' is following '\\' when preparing filters.  in some scenearios like objectGuid 294cf255-828c-43fd-8e61-36315c2b6237 the byte array legitimately has '\\' as a valid byte that is not hex formatted value

example filter (objectGuid=294cf255-828c-43fd-8e61-36315c2b6237)
the prepared filter value should be 0x55F24C298C82FD438E6136315C2B6237 
but instead is 0x55f24c298c82fd438e6136310b37


sonicsqwirl:
broke filters in the process...

PR #564, opened 2018-07-31
doc: fix SASL reference in bind example

ktdreyer:
The first line of this example imports the `SASL` value directly from ldap3, so we should not refer to this as `ldap3.SASL` in the code later on.


cannatag:
Thanks, I've updated docs manually upon your suggestion. So I close this PR withoute merging.

 Bye,
Giovanni

PR #566, opened 2018-08-06
Support formatting Active Directory time deltas

mprahl:
There are a few LDAP attributes that are stored on an Active Directory domain object that are negative "file time" values that represent time deltas. This PR formats these to Python time deltas for convenience.

I couldn't verify that the tests actually pass though because when I run the tests locally, I get the message `ERROR: Failure: Exception (testing location UNKNOWN-EDIR is not valid)`.

### DIFF
--- a/ldap3/protocol/formatters/formatters.py
+++ b/ldap3/protocol/formatters/formatters.py
@@ -324,6 +324,19 @@ except ImportError:
         return raw_value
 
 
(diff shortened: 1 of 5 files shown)


cannatag:
Please, could you send the PR to the "dev" branch?

mprahl:
@cannatag resubmitted as part of #573

PR #568, opened 2018-08-08
Wrap response_name in OctetString in mock_extended

Busuwe:
With the changes to ldap3/operation/extended.py in 6698d591d2a7b6efb961e916d14e0c5ae99ee198, introduced in 2.5.1, mock_extended in mockBase.py can no longer return the response_name as a simple string, and must return it as a OctetString instead.


cannatag:
Thanks!

PR #576, opened 2018-08-19
Fix WhoAmI calls using the mock LDAP instance

mprahl:
The `extended_response_to_dict` function expects the `responseValue` to be an `OctetString` but before this change, the mock LDAP instance would return a normal string and the `extended_response_to_dict` function would fail when trying to call `hasValue` on the string.


mprahl:
@cannatag could you please review?

PR #591, opened 2018-09-25
Prevented unnecessary raising of an error.

johnnysheppard-isode:
- The comment above the code suggested that an error should only be raised if: `raise_exception == True`, however an exception was raised whether **raise_exception** was _True_ or _False_.


cannatag:
Thank you. It will be included in the next release of ldap3

PR #605, opened 2018-10-26
add store= to Connection

clhedrick:
See issue. Note that I have tested this patch with the current version from PIP. I haven't actually built from this source and tested the patch. I'm not a python person, and I have no idea how to build the source and install it for testing.

If you'd prefer it for master, I can generate a pull request for it as well.


cannatag:
thanks, I don't use kerberos, so I cannot test your patch. I will include in in the next release of ldap3.

cannatag:
I've changed the "store" parameter to "cred_store"

PR #608, opened 2018-10-29
issue 607 ldifProducerStrategy breaks on trying to add a key value pa…

antoinell:
The following change should fix a runtime error in ldifProducerStrategy

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,5 @@ target/
 
 #PyCharm
 .idea/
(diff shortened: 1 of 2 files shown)


cannatag:
thanks. I will add this pr manually to the dev branch

antoinell:
Thanks Giovanni ! Thanks for creating and maintaining the ldap3 library !

Antoine

Issue #603, opened 2018-10-26
LDAPSASLBindInProgressError forever raises with NTLM


ribx:
I guess quite the same issue as with #227. Is there also a finally block needed?

https://github.com/cannatag/ldap3/blob/ab071f521801fced6f7e650a0ecc4e85ff362b34/ldap3/core/connection.py#L1270-L1314

The error is not easily reproducible and I am using a flask plugin with catches all exceptions, so I am not 100% sure, if we have the same problem here, as I cannot find any other exception in the logs.

ribx:
I managed to reproduce the error: When an invalid user is given, the area above is left with an `LDAPInvalidCredentialsResult` exception. The `LDAPSASLBindInProgressError` is then thrown when `connection.unbind()` is called.

cannatag:
This will be fixed in the next release of ldap3.

Issue #227, opened 2016-10-21
LDAPSASLBindInProgressError forever raises after network error


styleex:
https://github.com/cannatag/ldap3/blob/master/ldap3/core/connection.py#L1072

If a network error during the execution of the functions of authorization, the variable is not set to false. After that, the client always gets exception:

      Traceback (most recent call last):
        File "/usr/lib/python3.5/dist-packages/my_package/application.py", line 201, in rpc_import_security_groups
          ldap_result = await ldap.query_security_groups(domain.domain)
        File "/usr/lib/python3.5/dist-packages/my_package/ldap.py", line 126, in query_security_groups
          conn = await self.get_connection(domain)
        File "/usr/lib/python3.5/dist-packages/my_package/ldap.py", line 122, in get_connection
          conn.bind()
        File "/usr/lib/python3.5/dist-packages/ldap3/core/connection.py", line 535, in bind
          self.refresh_server_info()
        File "/usr/lib/python3.5/dist-packages/ldap3/core/connection.py", line 1125, in refresh_server_info
          self.server.get_info_from_server(self)
        File "/usr/lib/python3.5/dist-packages/ldap3/core/server.py", line 423, in get_info_from_server
          self._get_dsa_info(connection)
        File "/usr/lib/python3.5/dist-packages/ldap3/core/server.py", line 349, in _get_dsa_info
          get_operational_attributes=True)
        File "/usr/lib/python3.5/dist-packages/ldap3/core/connection.py", line 667, in search
          response = self.post_send_search(self.send('searchRequest', request, controls))
        File "/usr/lib/python3.5/dist-packages/ldap3/strategy/base.py", line 283, in send
          raise LDAPSASLBindInProgressError(self.connection.last_error)
      ldap3.core.exceptions.LDAPSASLBindInProgressError: cannot send operation requests while SASL bind is in progress

Need use try..finally block:

    self.sasl_in_progress = True
    try:
        if self.sasl_mechanism == EXTERNAL:
            result = sasl_external(self, controls)
        elif self.sasl_mechanism == DIGEST_MD5:
            result = sasl_digest_md5(self, controls)
        elif self.sasl_mechanism == GSSAPI:
            from ..protocol.sasl.kerberos import sasl_gssapi  # needs the gssapi package
            result = sasl_gssapi(self, controls)
    finally:
        self.sasl_in_progress = False

cannatag:
[...]

PR #609, opened 2018-10-30
Set `sasl_in_progess` to false in `do_ntlm_bind()` if exception is raised.

ribx:
encapsulate `do_sasl_bind()` in try-finally block and set `sasl_in_progess` always to false also if an exception is raised. see #227 and #603.

The diff is a bit confusing here. The only thing I did, was indenting the whole section into a try-finally block.


cannatag:
Manually added to dev branch. thanks.

PR #616, opened 2018-11-19
strategy/reusable: correctly refresh server info

krisztian-kovacs:
It turns out that Connection._fire_deferred() is a no-op for non-lazy
connections. Because of this the pooled connections of ReusableStrategy
do not correctly get their schema refreshed when
Connection.refresh_server_info() is called on a connection with reusable
strategy.

This causes performing operations without the schema being present if we're
using ServerPool() with multiple servers and leads to unexpected types
returned by Connection.search().

References: jira PAM-7321

### DIFF
--- a/ldap3/strategy/reusable.py
+++ b/ldap3/strategy/reusable.py
@@ -249,7 +249,7 @@ class ReusableStrategy(BaseStrategy):
                                     if pool.tls_pool and not self.worker.connection.tls_started:
                                         self.worker.connection.start_tls(read_server_info=False)
                                 if self.worker.get_info_from_server and counter:
-                                    self.worker.connection._fire_deferred()
+                                    self.worker.connection.refresh_server_info()
                                     self.worker.get_info_from_server = False
                                 response = None
                                 result = None


cannatag:
thanks! I include your fix in the dev branch. It will be available in the next release of ldap3.

krisztian-kovacs:
Giovanni, is there a good reason why you've merged this change via a non-standard and very hard to follow way?

Just asking because I was checking the current dev branch and seeing a commit message 'multiple fixes from github (PRs sent to master instead of dev)' makes me cry. It basically makes the git history untrackable, blame won't work and there's no way to find out later why some changes were done and what exactly was fixed.

This way we lose the original commit messages, too, which -- hopefully -- contain useful information...

cannatag:
Hi Krisztián, The only reason is that I was in a hurry to make some tests to the dev branch and I wanted to include all the PR from master. You’re right, this is not the correct way to handle them but otherwise I should had ask to every contributor to send the pr again to the dev branch. I understand that a cumulative commit is a very bad practice. If you think this should not be done  I can revert the PR and commit each single PR.

Thanks for your support.
Giovanni

PR #618, opened 2018-11-29
core/pooling: use a namedtuple to store server state

krisztian-kovacs:
Instead of constructing a list of three items to store the state of a
server in a pool, this change introduces the ServerState namedtuple.

This has the advantage that parts of the tuple can be referenced by name,
so instead of writing servers[offset][0] we can now write
server_states[offset].server.


cannatag:
thanks. Will include it manually in the dev branch,

PR #619, opened 2018-11-30
Fix error with unsafe distinguishedName in ad_add_members_to_groups function

MaximZemskov:
Fix bug with unsafe dn in ad_add_members_to_groups function


cannatag:
Thanks, I've manually updated the "dev" branch.

PR #645, opened 2019-01-15
Corrected object_class search filter

itbill:
The "object_class" attribute in search filter in search example should be "objectClass" instead.


cannatag:
thanks!

Issue #639, opened 2019-01-08
format_ad_timedelta throws exception


dirkjanm:
v2.5.2 seems to break on both python 2 and 3 with the new formatter for AD timedelta objects:

```
File "/home/dirkjan/.local/share/virtualenvs/venv/local/lib/python2.7/site-packages/ldap3/operation/search.py", line 453, in checked_attributes_to_dict_fast
    checked_attributes[name] = format_attribute_values(schema, name, decode_raw_vals_fast(attribute[3][1][3]) or [], custom_formatter)
  File "/home/dirkjan/.local/share/virtualenvs/venv/local/lib/python2.7/site-packages/ldap3/protocol/formatters/standard.py", line 213, in format_attribute_values
    formatted_values = [formatter(raw_value) for raw_value in values]  # executes formatter
  File "/home/dirkjan/.local/share/virtualenvs/venv/local/lib/python2.7/site-packages/ldap3/protocol/formatters/formatters.py", line 338, in format_ad_timedelta
    return format_ad_timestamp(raw_value * -1) - format_ad_timestamp(0)
TypeError: unsupported operand type(s) for -: 'str' and 'datetime.datetime'
```

In the call stack `raw_value` is a bytes object, which is multiplied with -1 (an integer). This will not produce a valid result and thus the `format_ad_timestamp` function will error out and return a bytes object, from which a `datetime.datetime` object is then substracted, which raises this exception.

In a related note, adding this filter changes the output for some objects from an int object to a datetime.datetime object (which is a bit unexpected for a patch release), which breaks tools that use the library and expect an int object.

cannatag:
HI, this code was submitted by a user, I cannot check it with AD. Do you mean that it always fails?

dirkjanm:
Yes it always fails. Seems like the user who submitted it did not test it properly.

yaap7:
Same issue here using v2.5.2 with python3.

```
[long error stack]
  File "/home/yaap/.local/lib/python3.5/site-packages/ldap3/protocol/formatters/standard.py", line 213, in format_attribute_values
    formatted_values = [formatter(raw_value) for raw_value in values]  # executes formatter
  File "/home/yaap/.local/lib/python3.5/site-packages/ldap3/protocol/formatters/standard.py", line 213, in <listcomp>
    formatted_values = [formatter(raw_value) for raw_value in values]  # executes formatter
[...]

PR #660, opened 2019-03-17
Fix issue #639

yaap7:
The issue #639 seems to be fixed with this little patch.

I think @may-day have the correct idea, but since the `format_ad_timestamp` uses a bytes as argument, the negative number should be corrected in this function instead of `format_ad_timedelta`.

Successfully tested with python 3.7.0 and python 2.7.12 from Ubuntu and an LDAP server which is running Windows 2012 R2.

### DIFF
--- a/ldap3/protocol/formatters/formatters.py
+++ b/ldap3/protocol/formatters/formatters.py
@@ -324,6 +324,24 @@ except ImportError:
         return raw_value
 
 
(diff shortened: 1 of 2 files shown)


cannatag:
Thanks, Can't merge this to master. Will add manually to dev and this fix will be released in the next version of ldap3.

PR #675, opened 2019-04-06
add json converter for timedelta

dirkjanm:
The json encoder doesn't like timedelta objects that were added for some AD properties in 2.5.2, encoding them as a string like datetime objects fixes this.


cannatag:
thanks!

PR #718, opened 2019-07-24
EntryState: `entry_raw_attributes` is populated instead of `raw_attributes`

cfelder:
According to the constructor of `EntryState` the variable name should be
`raw_attributes` whereas the rest of the code uses `entry_raw_attributes`
for populating the dictionary resulting in an empty `raw_attributes` dict.
The `Entry` class uses a property `entry_raw_attributes` for providing access.

For backwards compatibility with possibly external software using `EntryState` the previous name is provided as property.


cannatag:
thanks!

PR #724, opened 2019-08-08
Test on Python 3.7 and PyPy3

johnthagen:
`dist: xenial` is required for Python 3.7.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,13 @@
 language: python
+dist: xenial
 python:
   - "2.6"
   - "2.7"
   - "3.6"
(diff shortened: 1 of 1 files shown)


johnthagen:
Hmm, looking at the [commit history](https://github.com/cannatag/ldap3/commits/master), it looks like the CI is always failing?

cannatag:
Hi, sorry for this but the lab of different servers that I used for testing is not working anymore. I had a subscription on Azure but it expired. I test only locally on virtual machines. The travis configuration still works but it always fails.

Bye,
Giovanni

PR #731, opened 2019-08-27
Return offset timezone aware datetime for too large AD timestamp

jussih:
The function format_ad_timestamp normally returns offset aware
timestamp. The fallback value needs to be of the same format for
datetime comparisons to work.


cannatag:
Thanks! Will include it in the next ldap3 release.

PR #733, opened 2019-08-28
Accept 0 as valid input for pwdLastSet

RincewindsHat:
0 is a valid value for pwdLastSet, it will configure a Windows Active
Directory to accept the password only once and force the user to choose
a new one


cannatag:
Thanks, I will include the change in the dev code. Please next time do not commit to the "master" branch, use the "dev" branch instead. 

I will include this fix in the next release of ldap3

PR #742, opened 2019-09-28
Indentation for LDAPCertificateError during hostname check

ardayavas:
Modified indentation in file ldap3/core/tls.py at line 321

Hostname check was failing regardless of hostname match.


ardayavas:
Thank you, valid_names works just file.

PR #768, opened 2019-12-08
Added support for source specifications in LDAP connections

zorn96:
Added support for specifying a source address and source port or source port list when opening an LDAP connection.

In many simple environments, just opening a connection to a remote server at some destination address
and port is all you need.
However, sometimes you're working an environment with multiple network partitions which aren't mutually
routable. Or you're dealing with a firewall that only allows connections through if they originate from
a specific source address or address range, and you're on a machine with multiple identities from a
networking perspective.
Alternatively, you may have multiple network identities, and you're doing mutual TLS so you want to make
sure you pick the correct one. Or you just want to control source network identity at the application
level, rather than having it be controlled via ip tables or network switch configurations. Maybe an
application has a designated range of local ports that it owns, and you want to confine it to that
range within the application rather than by using a container.

Regardless of the specific reason, there's a number of use cases for specifying a source ip/hostname to
try to use when binding the socket for a connection, as well as for specifying a source port/port list
for binding the socket locally.
This change adds support for specifying such a source address, a source port, or source port list.

The idea with this change is that source address/port are attributes of the connection, rather than the
server definition. However, to faciliate checking server availability (which is done via the server
object) with this functionality, a source address, source port, and source port list can be specified
when checking server availability as well via optional arguments.

### DIFF
--- a/ldap3/core/connection.py
+++ b/ldap3/core/connection.py
@@ -65,11 +65,13 @@ from .usage import ConnectionUsage
 from .tls import Tls
 from .exceptions import LDAPUnknownStrategyError, LDAPBindError, LDAPUnknownAuthenticationMethodError, \
     LDAPSASLMechanismNotSupportedError, LDAPObjectClassError, LDAPConnectionIsReadOnlyError, LDAPChangeError, LDAPExceptionError, \
-    LDAPObjectError, LDAPSocketReceiveError, LDAPAttributeError, LDAPInvalidValueError, LDAPConfigurationError
+    LDAPObjectError, LDAPSocketReceiveError, LDAPAttributeError, LDAPInvalidValueError, LDAPConfigurationError, \
+    LDAPInvalidPortError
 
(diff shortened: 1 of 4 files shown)


zorn96:
hi @cannatag ! thanks so much for all your work maintaining and guiding the development of this library. I've found it incredibly helpful over the years.

Recently I was playing around with it and encountered some weird issues that I later discovered were due to firewall rules on the network, as the default routing on my linux machine was picking a source address/port that didn't always work.

I wrote up a monkeypatch to workaround it by modifying the socket before binding, but figured that I'd turn it into a cleaner, generally useful change and contribute it back for others to use. So that's what this PR is.
Let me know if you have any questions/if this seems unreasonable :)

zorn96:
Oh I should probably submit this to the dev branch shouldn't I

cannatag:
This is awesome, thanks!

Will try to include it in the next release of ldap3.

Thanks, 
Giovanni

cannatag:
Can you check the code in dev? Thanks

PR #779, opened 2020-01-20
Add Tls.ssl_options param.

Nazarii:
Currently there is not option to set custom ssl options when using Tls, only exact protocol version could be set.
After the change we could set Tls(version=ssl.PROTOCOL_SSLv23,
                                  ssl_options=[ssl.OP_NO_SSLv2,
                                               ssl.OP_NO_SSLv3,
                                               ssl.OP_NO_TLSv1,
                                               ssl.OP_NO_TLSv1_1])
This will allow to set the minumin TLSv1_2 version to use but not hte exact.


cannatag:
Hi, could you send the pull request to the "dev" repository?

Thanks,
Giovanni

cannatag:
Thanks, will include this in the next release of ldap3.

PR #781, opened 2020-01-31
If there are a Operational Attribute, the operation attr.virtual return error

ruytakata:
If there are a customized attribute with ';' character, it is considered a Operational Attribute, and the attr.virtual operation return error. Using hasattr(attr,'virtual') there is no error


cannatag:
thanks! Will be included in the next release of ldap3

PR #782, opened 2020-02-03
Introduce threading.Event to replace loop checking in async strategy

krizex:
The current implementation of the ASYNC strategy checks the response every 0.05s
by default and it is a configurable item.
The mechanism is inefficient so introduce a event base checker.

Brief of the implementation:
1. Add a new dict to store the (message_id, event) pair
2. Add a new lock for the above dict
3. Create a Event object for every message_id
4. Set the event when receiving the entire response of the message
5. Wait on the event when accessing the response

Signed-off-by: Yang Qian <yang.qian@citrix.com>

### DIFF
--- a/ldap3/strategy/asynchronous.py
+++ b/ldap3/strategy/asynchronous.py
@@ -23,7 +23,7 @@
 # along with ldap3 in the COPYING and COPYING.LESSER files.
 # If not, see <http://www.gnu.org/licenses/>.
 
-from threading import Thread, Lock
+from threading import Thread, Lock, Event
 import socket
 
 from .. import get_config_parameter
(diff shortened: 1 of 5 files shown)


krizex:
@cannatag Please have a look when you get time, thanks in advance!

cannatag:
Thanks, will work on this in the weekend.

cannatag:
Thanks, will test it and will include in the next release of ldap3.

krizex:
Thanks. Looking forward to see it.

cannatag:
You can test it in the dev branch.

krizex:
Yes, I have already tested in my private branch.

Issue #800, opened 2020-03-13
ImportError: cannot import name PyAsn1UnicodeDecodeError


qq719862911:
What should I do
In Ubuntu 16.04, python2.7.5, install from the source, and then execute 
"
#coding=utf-8
from ldap3 import Server, Connection, NTLM
def test():
    server1 = Server("***.**.com", port=636, use_ssl=True, get_info='ALL', connect_timeout=5)
    conn = Connection(
        server = server1,
        auto_bind = True,
        authentication = NTLM,
        read_only = False,
        user = 'test\\****',
        password = "!*****",
        fast_decoder=True,
        check_names=True,)
    return conn.user
if __name__ == '__main__':
    print(test())

"
 in python2.5.7 to report an error: 
Traceback (most recent call last):
  File "simpleforldap3.py", line 2, in <module>
    from ldap3 import Server, Connection, NTLM
  File "build/bdist.linux-x86_64/egg/ldap3/__init__.py", line 139, in <module>
  File "build/bdist.linux-x86_64/egg/ldap3/core/connection.py", line 44, in <module>
  File "build/bdist.linux-x86_64/egg/ldap3/operation/compare.py", line 28, in <module>
  File "build/bdist.linux-x86_64/egg/ldap3/operation/search.py", line 41, in <module>
ImportError: cannot import name PyAsn1UnicodeDecodeError

dirickxmaarten:
+1 

CentOS 8.1
pip install ldap3 works fine
after that I get:
```
[root@myhost scripts]# python3
Python 3.6.8 (default, Nov 21 2019, 19:31:34) 
[GCC 8.3.1 20190507 (Red Hat 8.3.1-4)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import ldap3
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.6/site-packages/ldap3/__init__.py", line 139, in <module>
    from .core.connection import Connection
  File "/usr/local/lib/python3.6/site-packages/ldap3/core/connection.py", line 44, in <module>
    from ..operation.compare import compare_operation, compare_request_to_dict
  File "/usr/local/lib/python3.6/site-packages/ldap3/operation/compare.py", line 28, in <module>
    from ..operation.search import ava_to_dict
  File "/usr/local/lib/python3.6/site-packages/ldap3/operation/search.py", line 41, in <module>
[...]

PR #802, opened 2020-03-15
Fixes issue #800 - update requirements.txt to a later version of pyasn1

jannh:


clokep:
This looks like the right fix to me! Without this you can run into a broken ldap3 install. Any chance of getting this merged / having a point release done? Thanks!

dylancis:
Hi, would be great to have to merged. I still have the issue.

cannatag:
Hi, It should be at least 0.3.1. Will modify the requirements.txt by myself. thanks!

cannatag:
You're right, but I will include it in dev. Thanks!

PR #805, opened 2020-03-16
Dont throw away operational attributes if explicitly requested

Busuwe:
If an operational attribute is requested explicitly (as opposed to using '+') using a mocked server, then keep it in the responses


cannatag:
thanks!

PR #820, opened 2020-05-04
Update pyasn1 version.

Nazarii:
`PyAsn1UnicodeDecodeError`' is added  in pyasn1 V0.4.6, we will receive ImportError on ldap3 import when using older pyasn1 versions


cannatag:
thanks, will include in dev.

PR #821, opened 2020-05-06
Use ISO8601 formatting for JSON date-time values

hughpyle:
Python's default str(datetime) uses a space to separate the date and time portions.  ISO8601 requires a "T" separator.

JSON doesn't mandate any particular datetime format, but the ISO standard format is widely understood, and some systems won't accept date-times that do not use the ISO format (including the "T").

With this change:  LDAP objects formatted with to_json() use ISO8601 formatting in their date-time values.

### DIFF
--- a/ldap3/utils/conv.py
+++ b/ldap3/utils/conv.py
@@ -188,7 +188,7 @@ def format_json(obj):
         return obj._store
 
     if isinstance(obj, datetime.datetime):
-        return str(obj)
+        return obj.isoformat()
 
     if isinstance(obj, int):
         return obj


cannatag:
Hi, I think that this would break existing code. I will try make it optional with something like to_json(iso_date=True) or similar.

Thanks,
Giovanni

cannatag:
Added iso_format parameter, defaults to False.

Thanks Hugh. It will be included in the next release of ldap3.

hughpyle:
Great, thanks Giovanni!

PR #822, opened 2020-05-07
Always use utf-8 when reading JSON entries

sscherfke:
On Windows, `open()` does not necessarily use utf-8 but JSON must always be utf-8, so we can safely enforce it when opening an entries JSON file.


cannatag:
please can you send the merge request to dev? Thanks

PR #835, opened 2020-06-29
Fix lost error message during do_auto_bind()

cfelder:
* self.unbind() implicitly sets ``self.last_error = None``
* store error message on local variable and use this as argument for
  LDAPBindError


cannatag:
thanks, I will include it in "dev" branch.

Issue #831, opened 2020-06-04
Fail to connect Windows DC which LdapEnforceChannelBinding is 2


krizex:
After applying the security update from Microsoft https://support.microsoft.com/en-us/help/4034879/how-to-add-the-ldapenforcechannelbinding-registry-entry, ldap3 is unable to connect the DC with NTLM or Kerberos.

Get the following error:
```
  File "/usr/local/lib/python3.6/dist-packages/ldap3/core/connection.py", line 587, in bind
    response = self.do_sasl_bind(controls)
  File "/usr/local/lib/python3.6/dist-packages/ldap3/core/connection.py", line 1293, in do_sasl_bind
    result = sasl_gssapi(self, controls)
  File "/usr/local/lib/python3.6/dist-packages/ldap3/protocol/sasl/kerberos.py", line 94, in sasl_gssapi
    result = send_sasl_negotiation(connection, controls, out_token)
  File "/usr/local/lib/python3.6/dist-packages/ldap3/protocol/sasl/sasl.py", line 160, in send_sasl_negotiation
    response = connection.post_send_single_response(connection.send('bindRequest', request, controls))
  File "/usr/local/lib/python3.6/dist-packages/ldap3/strategy/sync.py", line 121, in post_send_single_response
    responses, result = self.get_response(message_id)
  File "/usr/local/lib/python3.6/dist-packages/ldap3/strategy/base.py", line 401, in get_response
    raise LDAPOperationResult(result=result['result'], description=result['description'], dn=result['dn'], message=result['message'], response_type=result['type'])
ldap3.core.exceptions.LDAPInvalidCredentialsResult: LDAPInvalidCredentialsResult - 49 - invalidCredentials - None - 80090346: LdapErr: DSID-0C090579, comment: AcceptSecurityContext error, data 80090346, v3839 - bindResponse - None
```

cannatag:
Please resend the latest merge request to the "dev" branch. Thanks

cannatag:
Will be included in the next released of ldap3

PR #838, opened 2020-07-06
Add channel binding support for SASL + TLS

krizex:
When building the connection base on SASL(Kerberos) + TLS to connect a Windows
ldap server (aka Domain Controller), we have to enforce ldap channel binding for
the connection otherwise the server will reject the connection.
See: https://support.microsoft.com/en-us/help/4520412/2020-ldap-channel-binding-and-ldap-signing-requirements-for-windows

This change also consider the useage of the SASL without TLS connection, so
won't calculate the channel binding token for that case.

Signed-off-by: Yang Qian <yang.qian@citrix.com>


krizex:
This is a fix for #831

krizex:
Close #832 and create this one targeting dev branch instead.

cannatag:
Thanks for being so fast!

Issue #612, opened 2018-11-05
invalid escape sequence warning


cmclaughlin:
Running pytest against my project that uses ldap3 shows this warning:

```

========================================= warnings summary =========================================
.eggs/ldap3-2.5.1-py3.6.egg/ldap3/protocol/formatters/validators.py:355
  xxxxx/.eggs/ldap3-2.5.1-py3.6.egg/ldap3/protocol/formatters/validators.py:355: DeprecationWarning: invalid escape sequence \9
  xxxxx/.eggs/ldap3-2.5.1-py3.6.egg/ldap3/protocol/formatters/validators.py:355: DeprecationWarning: invalid escape sequence \9

```

cannatag:
Hi, can you send the code of your test? I cannot replicate the issue.

cannatag:
I'm closing this issue. Please reopen if you still get this problem.

cmclaughlin:
Sorry, I must have been doing something silly... at some point in my project the warning went away.

PR #841, opened 2020-07-22
Fix "invalid escape sequence \9" warning

stevenwinfield:
Mark validate_uuid_le's docstring as a raw string, otherwise "\9" et al. are interpreted as escape sequences - some invalid.

Fixes https://github.com/cannatag/ldap3/issues/612, which I believe couldn't be reproduced because the warning is only raised when the bytecode for this module is first generated. After that, the problem would seem to disappear.

I still see warning in v2.7, though.


cannatag:
thanks. I've updated the dev branch with your fix. I must reject the merge in the master branch.

Bye,
Giovanni

PR #842, opened 2020-07-24
Don't use Markdown syntax in rST text

mcepl:


cannatag:
Thanks for the fix. I've manually updated the source code in the dev branch. Please don't send merges to the master branch.

Issue #115, opened 2015-10-12
Mock LDAP server for testing?


midnightlynx:
Projects using `python-ldap` or `pyldap` for their LDAP connection can use the `mockldap` package in unit testing, so as not to make any changes in real LDAP server data.

Do you know if `mockldap` is compatible with `ldap3`-based unit testing, and if not, do you know if any projects of that sort exist that are?

cannatag:
Hi, I've been working on the MockSync and MockAsync strategies inside ldap3 for a while, but I didn't go much farther because I had not real need for them. If  you are interested I can try to keep working on them. My idea was to emulate an ldap server with default responses. Do you have any suggestion on how these strategies should work? 

Bye,
Giovanni

-------- Messaggio originale --------
Da:midnightlynx <notifications@github.com>
Inviato:Tue, 13 Oct 2015 01:37:52 +0200
A:cannatag/ldap3 <ldap3@noreply.github.com>
Oggetto:[ldap3] Mock LDAP server for testing? (#115)

midnightlynx:
The general approach in `mockldap` is for the developer to provide an initial mock LDAP directory in dictionary form that queries are run against. Reading queries return the appropriate data from the dictionary, writing queries remove or modify what's in the dictionary, etc.
The mock LDAP object internally records which method calls are made, with which arguments, and provides a method to inspect this record.
They haven't fully implemented mocking of all query operations against the mock dictionary, so they also provide a way to seed methods so that a particular given set of args and kwargs will return a particular seeded result.
The way they implemented the recording and seeding can be here, and is independent of their particular implementation of the LDAP interaction:
https://bitbucket.org/psagers/mockldap/src/b6b3fba5345e04b511b947c2184c501d37a5156c/src/mockldap/recording.py

So... Maybe something like that?

cannatag:
My idea with MockSync and MockAsync is to describe in a json file what a specific operation is expected to return, so this is similar to the recording/seeding option in mockldap. I also started a project to build a pure python ldap 3 server bit this is still very rudimentary. Once it becomes more functional it could bé used forse mocking too. If you're interested I could tra to speed up its development.

midnightlynx:
Please see what you can do, if it's not too much trouble. I could find use for them.

cornelinux:
Hi I created my own simple ldap3mock:
https://github.com/privacyidea/privacyidea/blob/master/tests/ldap3mock.py

which I am using here
[...]

PR #843, opened 2020-07-24
Don't crash whole test suite with errors, just skip over missing LDAP server

mcepl:
Hi,

I am OpenSUSE maintainer of many Python packages and we try to run unit test in as many of our packages as possible. However, obviously we don’t have true LDAP server installed in the build system, the test suite crashes hard (because the function mistakenly uses `Exception()` instead of at least `AssertionError()`, which would make tests just fail).

Even better is just throw `unittest.SkipTest()` which makes failing tests be skipped in such build system.

I see on #115 that there are some mocking facilities inside of this package, is it possible somehow easily run the test suite against the mocked server?

Also, I am on the crusade of eliminating nose1 from our distribution. When I look at its repository on https://github.com/nose-devs/nose, the last release 1.3.7 was on 2 Jun 2015, and even the last commit on the master branch was on 4 Mar 2016. It seems to me that this project uses nose only for its test runner and it can be just replaced by for example pytest runner (or even ``python3 -munittest``).

### DIFF
--- a/test/config.py
+++ b/test/config.py
@@ -25,6 +25,7 @@ from os import environ, remove
 from os import path
 from random import SystemRandom
 from tempfile import gettempdir
+from unittest import SkipTest
 
 from ldap3 import SIMPLE, SYNC, ROUND_ROBIN, IP_V6_PREFERRED, IP_SYSTEM_DEFAULT, Server, Connection,\
     ServerPool, SASL, STRING_TYPES, get_config_parameter, set_config_parameter, \
@@ -380,7 +381,7 @@ elif location == 'W10GC9227-AD':
     test_logging_filename = path.join(gettempdir(), 'ldap3.log')
     test_valid_names = ['10.160.201.232']
 else:
-    raise Exception('testing location ' + location + ' is not valid')
+    raise SkipTest('testing location ' + location + ' is not valid')
 
 if test_logging:
     try:


cannatag:
Hi, sorry for the late answer! I am an “historical” user of SuSE products and openSUSE distributions. I agree with you that testS should be done whenever possible. Some tests in the ldap3 library are for testing the library itself, but other need a real LDAP server because all the low-level network communication is done in the source code. The mocking part of the library is for developers that want to test code that uses the ldap3 library. I’m not using nose anymore, it can safely removed from the project.

Thanks for your support,
Giovanni

mcepl:
@cannatag So, what’s missing from this PR to be merged?

cannatag:
I work on this project in my spare time, usually on nighttime and on weekend.  Will work on it soon!

mcepl:
> I work on this project in my spare time, usually on nighttime and on weekend. Will work on it soon!
> […](#)
> Il giorno 31 lug 2020, alle ore 09:22, Matěj Cepl ***@***.***> ha scritto: ﻿ @cannatag So, what’s missing from this PR to be merged? — You are receiving this because you were mentioned. Reply to this email directly, view it on GitHub, or unsubscribe.

I didn’t want to rush you, just curious why you wrote generally positive review, without hitting that [Merge] button.

cannatag:
Thanks for the fix. I've manually updated the source code in the dev branch. Please don't send merges to the master branch.

mcepl:
> Thanks for the fix. I've manually updated the source code in the dev branch. Please don't send merges to the master branch.

Then make the dev branch default in your settings to make it obvious where you want your contributions.

cannatag:
That's a good idea. Done.

Thanks again,
Giovanni

cannatag:
Hi, I've release the 2.8 version of ldap3. It includes a bunch of fixes and some new feature. Have a look at https://ldap3.readthedocs.io/en/latest/changelog.html for the changelog.

Bye,
Giovanni

PR #859, opened 2020-08-18
fixed attribute error in restartable class (Fix #858)

### LINKED ISSUE #858: RestartableStrategy object has no attribute 'unbind'
Version 2.8 doesn't work for me, I switched back to 2.7. Maybe its just the following error, but could also be an usage error...
```
ldap3.core.exceptions.LDAPMaximumRetriesError: LDAPMaximumRetriesError: restartable connection failed to send/Exception history:/    0 <class 'AttributeError'>: 'RestartableStrategy' object has no attribute 'unbind'/Maximum number of retries reached: 30
```
I think line https://github.com/cannatag/ldap3/blob/dev/ldap3/strategy/restartable.py#L156 should be `self.connection.unbind()`, shouldn't it?

cslux:
This bug was introduced in commit 191667cf.


cannatag:
thanks!

PR #895, opened 2020-12-03
Documentation updates to clean up old constants removed in v2.0.7

noritada:
Hello,

Some old constants, which were removed in v2.0.7, are found to be left behind in documents.
I have cleaned up them. Could you please check and apply?

Many thanks.


cannatag:
Thanks!

PR #898, opened 2020-12-21
Support MS Active Directory persistent search

elexpander:
Since LDAP persistent search was already implemented it was fairly easy to implement it for Active Directory copying the existing code.

AD persistent search has its own OID and fewer options: https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-adts/f14f3610-ee22-4d07-8a24-1bf1466cba5f

### DIFF
--- a/ldap3/extend/microsoft/persistentSearch.py
+++ b/ldap3/extend/microsoft/persistentSearch.py
@@ -1,0 +1,126 @@
+"""
+"""
+
+# Created on 2016.07.08
+#
+# Author: Giovanni Cannata
+#
+# Copyright 2016 - 2020 Giovanni Cannata
+#
+# This file is part of ldap3.
+#
+# ldap3 is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published
+# by the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# ldap3 is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with ldap3 in the COPYING and COPYING.LESSER files.
+# If not, see <http://www.gnu.org/licenses/>.
+
+try:
+    from queue import Empty
+except ImportError:  # Python 2
+    # noinspection PyUnresolvedReferences
+    from Queue import Empty
+
+from ...core.exceptions import LDAPExtensionError
+from ... import SEQUENCE_TYPES
+from ...utils.dn import safe_dn
+from ...protocol.microsoft import persistent_search_control
+
+
+class ADPersistentSearch(object):
+    def __init__(self,
+                 connection,
+                 search_base,
+                 search_filter,
+                 search_scope,
+                 dereference_aliases,
+                 attributes,
+                 size_limit,
+                 time_limit,
+                 streaming,
+                 callback
+                 ):
+        if connection.strategy.sync:
+            raise LDAPExtensionError('Persistent Search needs an asynchronous streaming connection')
+
+        if connection.check_names and search_base:
+            search_base = safe_dn(search_base)
+
+        self.connection = connection
+        self.message_id = None
+        self.base = search_base
+        self.filter = search_filter
+        self.scope = search_scope
+        self.dereference_aliases = dereference_aliases
+        self.attributes = attributes
+        self.size_limit = size_limit
+        self.time_limit = time_limit
+        self.connection.strategy.streaming = streaming
+        if callback and callable(callback):
+            self.connection.strategy.callback = callback
+        elif callback:
+            raise LDAPExtensionError('callback is not callable')
+
+        self.controls = [persistent_search_control()]
+
+        self.start()
+
+    def start(self):
+        if self.message_id:  # persistent search already started
+            return
+
+        if not self.connection.bound:
+            self.connection.bind()
+
+        with self.connection.strategy.async_lock:
+            self.message_id = self.connection.search(search_base=self.base,
+                                                     search_filter=self.filter,
+                                                     search_scope=self.scope,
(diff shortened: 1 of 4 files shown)


ldap3/extend/microsoft/persistentSearch.py
```diff
@@ -0,0 +1,125 @@
...
+        self.controls = [persistent_search_control()]
+
+        self.start()
+
+    def start(self):
+        if self.message_id:  # persistent search already started
+            return
+
+        if not self.connection.bound:
+            self.connection.bind()
```
azornberg-ps, line 80: does it make sense to implicitly bind here if the connection isn't using a lazy binding strategy? it also doesn't look like we're checking the result of the bind
should this operation log something if it needs to handle binding? or should it fail for an unbound connection? 

regardless of whether this operation implicitly binds or not, it seems like if we get past this point and the connection is not bound (e.g. we fail to bind here), we should raise an exception rather than continue onwards to take the lock and search

ldap3/extend/microsoft/persistentSearch.py
```diff
@@ -0,0 +1,125 @@
...
+        self.attributes = attributes
+        self.size_limit = size_limit
+        self.time_limit = time_limit
+        self.connection.strategy.streaming = streaming
+        if callback and callable(callback):
+            self.connection.strategy.callback = callback
+        elif callback:
+            raise LDAPExtensionError('callback is not callable')
+
+        self.controls = [persistent_search_control()]
```
azornberg-ps, line 71: can other controls be combined with a persistent search, around showing deleted objects and such? if so, it would be nice to support them being passed in by callers

elexpander: I don't think it supports any other controls. In fact, the AD persistent search is a change notification. It simply reports changes in existing AD objects.

If you want to be notified of object creation/deletion you may for instance run a persistent search for AD groups and look up the member attribute.

azornberg-ps: ah ok, thanks. I'm not familiar with persistent search as a mechanism, so the explanation helps. it might be nice to include use cases for this in the docstring

ldap3/extend/microsoft/persistentSearch.py
```diff
@@ -0,0 +1,125 @@
...
+
+        self.connection = connection
+        self.message_id = None
+        self.base = search_base
+        self.filter = search_filter
+        self.scope = search_scope
+        self.dereference_aliases = dereference_aliases
+        self.attributes = attributes
+        self.size_limit = size_limit
+        self.time_limit = time_limit
```
azornberg-ps, line 64: if AD persistent search is a change notification, then how do the time_limit and size_limit fields come into play?

similarly, how do attributes come into play? does attributes do the same thing, or does it constrain our search to only produce notifications on changes to the attributes specified?

elexpander: You are right, they don't. I have updated the code. Have also removed `dereference_aliases` since aliases are not even supported by AD.

`attributes` is a list of attributes that will be returned by notifications.

ldap3/extend/microsoft/persistentSearch.py
elexpander: You may be right here. This is just a copy-paste from the [standard peristent search method](https://github.com/cannatag/ldap3/blob/4d33166f0869b929f59c6e6825a1b9505eb99967/ldap3/extend/standard/PersistentSearch.py#L91). Whatever is done, the behaviour should be the same on both methods.

azornberg-ps: keeping the behavior consistent makes sense to me

ldap3/extend/microsoft/persistentSearch.py
```diff
@@ -0,0 +1,117 @@
...
+
+        self.connection = connection
+        self.message_id = None
+        self.base = search_base
+        self.scope = search_scope
+        self.attributes = attributes
+        self.controls = [persistent_search_control()]
+
+        # this is the only filter permitted by AD persistent search
+        self.filter = '(objectClass=*)'
```
azornberg-ps, line 60: minor nit: I think '(objectClass=*)' is available as a constant somewhere

also, since filter is a reserved word in python, it's probably not very python-y to use it as our class member. this also doesn't need to be a class member if its not changeable. it would make more sense as a constant declared in all caps like `FIXED_AD_PERSISTENT_SEARCH_FILTER = ...`

ldap3/extend/microsoft/persistentSearch.py
```diff
@@ -0,0 +1,117 @@
...
+    def next(self, block=False, timeout=None):
+        if not self.connection.strategy.streaming and not self.connection.strategy.callback:
+            try:
+                return self.connection.strategy.events.get(block, timeout)
+            except Empty:
+                return None
+
+        raise LDAPExtensionError('Persistent search is not accumulating events in queue')
+
+    def funnel(self, block=False, timeout=None):
```
azornberg-ps, line 105: is `funnel` a common term for this operation?
if not, could this function get a docstring

elexpander: This generator is copied from the [standard PersistentSearch](https://github.com/cannatag/ldap3/blob/4d33166f0869b929f59c6e6825a1b9505eb99967/ldap3/extend/standard/PersistentSearch.py#L125) and it does the same job in this case.

ldap3/extend/microsoft/persistentSearch.py
elexpander: doc added

ldap3/extend/microsoft/persistentSearch.py
elexpander: filter is not a reserved word and the string is not defined as constant

ldap3/extend/microsoft/persistentSearch.py
elexpander: doc added

cannatag:
Thanks! Do you think this is the final version?

elexpander:
Yeah, I don't plan any more changes.

Thanks for merging

PR #899, opened 2020-12-30
Implement LDAP signing (integrity checking) when using DIGEST-MD5 auth

### LINKED ISSUE #889: support LDAP signing (integrity checking) when using NTLM or DIGEST-MD5 auth
Hello, 

Since mars 2020, Microsoft Active Directory enables LDAP signing by default (see  https://support.microsoft.com/en-us/help/4520412/2020-ldap-channel-binding-and-ldap-signing-requirement-for-windows for details).  It is also possible (and recommended) to enforce request signing for LDAP requests on Microsoft AD (`LDAPServerIntegrity=2`). 

When LDAP signing is enforced:
- Simple authentication and NTLM authentication over SICILY are not allowed anymore
- MD5-DIGEST authentication over SASL is still allowed but with signing or encryption of each LDAP request.

The python `ldap3` module does not support integrity signing. It's an issue because it prevent ldap3 to perform queries on domains where this feature is enabled...

Any query made with `ldap3` when this feature is enabled will trigger the following error : `00002028: LdapErr: DSID-0C09027F, comment: The server requires binds to turn on integrity checking if SSL\\TLS are not already active on the connection, data 0`

Of course, one easy solution could be to connect using LDAPS / to start a StartTLS session (integrity signing is only mandatory when the connection is not encrypted) but it's not always desirable  : LDAPS/LDAP over StartTLS require some configuration (certificate generation & renewal) and may not be enabled for legitimate reasons on the MS domain controllers

Would it be possible to implement LDAP requests signing in ldap3? 

Since it would be quite complicated to rework entirely NTLM auth, I would propose to focus on MD5-DIGEST first.
From my understanding : 
- The part handling MD5-DIGEST authentication with integrity signing is already implemented in the lib (you just need to replace `qop = b'auth'` by `qop = b'auth-int'` in [digestMd5.py](https://github.com/cannatag/ldap3/blob/master/ldap3/protocol/sasl/digestMd5.py))
- However, the part handling the signature (or encryption) of subsequent messages is not. [As per the RFC](https://tools.ietf.org/html/draft-ietf-sasl-rfc2831bis-11#section-2.3), when `auth-int` is enabled then messages not having signature are purely and simply discarded by the server, so it's a little bit of a bummer .. 

The messages signature should probably be implemented directly in the [`sending()`](https://github.com/cannatag/ldap3/blob/master/ldap3/strategy/base.py#L868) function of each strategy ?

For information, the `cyrus-sasl-md5` package (which is used when performing `ldapsearch -Y DIGEST-MD5` or using `ldap-python` module) support LDAP signing. Its implementation for signing messages can be found [here](https://github.com/cyrusimap/cyrus-sasl/blob/master/plugins/digestmd5.c#L1491-L1591).

Related to  #848 , #860

Augustin-FL:
This PR implement optional LDAP signing (integrity protection) when using DIGEST-MD5 authentication.

LDAP signing is a way to prevent replay attacks & packets alteration attacks without encrypting the LDAP traffic. Microsoft publicly recommend to enforce signing when talking to an Active Directory server over a non-encrypted connection : https://support.microsoft.com/en-us/help/4520412/2020-ldap-channel-binding-and-ldap-signing-requirements-for-windows

My proposal : add an optional fifth argument to `sasl_credentials`: 

```
c = Connection(server, auto_bind = True, version = 3, client_strategy = test_strategy, authentication = SASL,
                             sasl_mechanism = DIGEST_MD5, sasl_credentials = (None, 'username', 'password', None, 'sign'))
```

- When this argument is set to the string `sign`, LDAP requests are signed and signature of LDAP responses is verified.
- When this argument is set to any other value or not set, LDAP requests are not signed.

This argument is ignored when not using DIGEST-MD5 authentication method. It is also ignored when using an LDAPS connection *(signing is not supported by most LDAP servers when using LDAPS, because the TLS connection is already providing replay protection so there is no need to sign)*

Fixes #889

### DIFF
--- a/docs/manual/source/bind.rst
+++ b/docs/manual/source/bind.rst
@@ -154,18 +154,28 @@ server trust the credential provided when establishing the secure channel::
 Digest-MD5
 ^^^^^^^^^^
 
-To use the DIGEST-MD5 mechanism you must pass a 4-value tuple as sasl_credentials: (realm, user, password, authz_id). You can pass None
-for 'realm' and 'authz_id' if not used. Quality of Protection is always 'auth'::
+To use the DIGEST-MD5 mechanism you must pass a 4-value or 5-value tuple as sasl_credentials: (realm, user, password, authz_id, enable_signing). You can pass None
+for 'realm', 'authz_id' and 'enable_signing' if not used::
 
     from ldap3 import Server, Connection, SASL, DIGEST_MD5
     server = Server(host = test_server, port = test_port)
     c = Connection(server, auto_bind = True, version = 3, client_strategy = test_strategy, authentication = SASL,
-                             sasl_mechanism = DIGEST_MD5, sasl_credentials = (None, 'username', 'password', None))
+                             sasl_mechanism = DIGEST_MD5, sasl_credentials = (None, 'username', 'password', None, 'sign'))
 
 Username is not required to be an LDAP entry, but it can be any identifier recognized by the server (i.e. email, principal, ...). If
 you pass None as 'realm' the default realm of the LDAP server will be used.
 
-**Again, remember that DIGEST-MD5 is deprecated and should not be used.**
+``enable_signing`` is an optional argument, which is only relevant for Digest-MD5 authentication. This argument enable or disable signing
+(Integrity protection) when performing LDAP queries.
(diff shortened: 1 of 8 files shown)


Augustin-FL:
Hello @cannatag ,

This PR is ready for review & possibly for merge.

cannatag:
Thanks!

PR #911, opened 2021-01-25
Fix typo in doc

lestercheung:
Thank you for maintaining the _ldap3_ module! It's so much easier that coming up LDIF files! ;-)

Just a drive-by typo fix - please review.


zorn96:
this seems like a small safe change. thanks for fixing the typo

PR #928, opened 2021-03-02
Added native kerberos support on windows using the winkerberos library.

zorn96:
This commit adds functionality for smart kerberos negotiation on windows and
posix systems. The library checks whether winkerberos or gssapi is installed,
and picks the appropriate kerberos negotiation accordingly.

All features supported with posix kerberos negotiation are supported with the
new native windows kerberos negotiation. This includes:
- Reverse DNS capabilities and target name specification
- Channel bind token support for AD
- Authentication only - no security layers

Much of the common logic has been split out so that improvements made to
one will be picked up by both.

### DIFF
--- a/ldap3/protocol/sasl/kerberos.py
+++ b/ldap3/protocol/sasl/kerberos.py
@@ -23,19 +23,31 @@
 # along with ldap3 in the COPYING and COPYING.LESSER files.
 # If not, see <http://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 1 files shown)


zorn96:
@cannatag I wanted your thoughts on this if you have the time.
I'm planning to add a bunch of units tests. basically everything that isn't platform-library-specific should be shareable and unit test-able i think

cannatag (approved):
Great, thanks Azaria!

PR #931, opened 2021-03-04
Documentation: python2 to python3 syntax change was added

paveldroo:
Previous code: `print entry['attributes']`
Changed to: `print(entry['attributes'])`


zorn96 (approved):

zorn96:
thanks for catching this!

PR #952, opened 2021-06-09
Fix doc example and clarify connection attribute

meesha7:
This fixes a small inconsistency in a code sample and clarifies the connection attribute doc.


meesha7:
Ah just noticed a similar fix in https://github.com/cannatag/ldap3/pull/945, feel free to close this one, whichever of these two is preferred :)

zorn96 (approved):

Issue #954, opened 2021-06-17
Socket never closed after certain exceptions, python warning displayed


HansWeltar:
Connecting to a non-existing port will later trigger a
`ResourceWarning: unclosed <socket.socket fd=3, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=6, laddr=('0.0.0.0', 58471)> ResourceWarning: Enable tracemalloc to get the object allocation traceback`

The behaviour can be reproduced as follows:
```python
import gc, ldap3, warnings
warnings.simplefilter('always') # enable warnings
conn = ldap3.Connection(ldap3.Server("127.0.0.1", port=1)) # nonexisting port
try: conn.bind()  # this will error due to the wrong port
except Exception: pass

del(conn); gc.collect() # force conn to be deleted to trigger the warning
```

We suspect the error might happen in strategy/base.py in _open_socket.
When an error happens in this function after the socket is created, the socket is not deleted, and the connection remains in state unbound, closed. Because of the inaccurate 'closed' state, the socket is never closed later in the lib.

Note that there are multiple exceptions being raised in _open_socket. All of them might potentially leave a socket behind.

PR #955, opened 2021-06-17
Cleanup socket when errors happen in _open_socket

HansWeltar:
Fixes Issue #954: Socket never closed after certain exceptions, python warning
displayed

Connecting to a non-existing port will later trigger a
ResourceWarning: unclosed <socket.socket fd=3, ....>

This warning occurs whenever an exception is raised from _open_socket, after
the socket was allocated. When an exception is raised from _open_socket,
the connection will remain in state unbound, closed, and so the allocated
socket will never be freed again.
Solution is to free the socket before raising the exception.

### DIFF
--- a/ldap3/strategy/base.py
+++ b/ldap3/strategy/base.py
@@ -194,6 +194,15 @@ class BaseStrategy(object):
         Tries to open and connect a socket to a Server
         raise LDAPExceptionError if unable to open or connect socket
         """
+        def _cleanup_socket():
+            """ Close the open socket. Call before raising exceptions so no
+                lingering sockets are left about.
+            """
+            try:
+                self.connection.socket.close()
+            except OSError:
+                pass
+
         try:
             self.connection.socket = socket.socket(*address[:3])
         except Exception as e:
@@ -231,6 +240,7 @@ class BaseStrategy(object):
                 if log_enabled(ERROR):
                     log(ERROR, 'Unable to locally bind to local address <%s> with any of the source ports <%s> for connection <%s due to <%s>',
                         self.connection.source_address, self.connection.source_port_list, self.connection, last_bind_exc)
+                _cleanup_socket()
                 raise communication_exception_factory(LDAPSocketOpenError, type(last_bind_exc)(str(last_bind_exc)))(last_bind_exc)
 
         try:  # set socket timeout for opening connection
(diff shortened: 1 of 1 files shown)


ldap3/strategy/base.py
```diff
@@ -194,6 +194,15 @@ def _open_socket(self, address, use_ssl=False, unix_socket=False):
...
         raise LDAPExceptionError if unable to open or connect socket
         """
+        def _cleanup_socket():
+            """ Close the open socket. Call before raising exceptions so no
+                lingering sockets are left about.
+            """
+            try:
+                self.connection.socket.close()
+            except OSError:
+                pass
```
zorn96, line 204: would this make more sense as an annotation to wrap functions than as a standalone function to be called? then it would check if an exception was being raised before deciding whether to close the socket or not

that way we get the niceties of python context management and it operates more like a try/finally for the entire function

HansWeltar: Is this what you meant? (i assumed you meant decorator, i'm not sure how to solve this with annotations)
Personally i'm not sure i find this much more readable.
```diff
diff --git a/ldap3/strategy/base.py b/ldap3/strategy/base.py
--- a/ldap3/strategy/base.py
+++ b/ldap3/strategy/base.py
@@ -23,6 +23,7 @@
 # along with ldap3 in the COPYING and COPYING.LESSER files.
 # If not, see <http://www.gnu.org/licenses/>.
 
+import functools
 import socket
 try:  # try to discover if unix sockets are available for LDAP over IPC (ldapi:// scheme)
     # noinspection PyUnresolvedReferences
@@ -189,6 +190,26 @@ class BaseStrategy(object):
         if self.connection.usage:
             self.connection._usage.stop()
 
+    def close_connection_on_exception(orig_func):
+        """
+        A decorator that will close self.connection.socket if the
+        original function errors. Should only be applied to functions that
+        are executed before self.connection.bound and self.connection.closed
+        are set (ie _open_socket)
+        """
+        @functools.wraps(orig_func)
+        def test_wrapper(self, *args, **kwargs):
+            try:
+                return orig_func(*args, **kwargs)
+            except Exception as e:
+                try:
+                    self.connection.socket.close()
+                except Exception:
+                    pass
+                raise
+        return test_wrapper
+
+    @close_connection_on_exception
     def _open_socket(self, address, use_ssl=False, unix_socket=False):
         """
         Tries to open and connect a socket to a Server
```

zorn96: yeah that's what i'd meant. i moreso meant that it was easier for people adding new things to realize that they should use it than easier to read what it does

ldap3/strategy/base.py
```diff
@@ -194,6 +194,15 @@ def _open_socket(self, address, use_ssl=False, unix_socket=False):
         Tries to open and connect a socket to a Server
         raise LDAPExceptionError if unable to open or connect socket
         """
+        def _cleanup_socket():
```
dolfinus, line 197: Why not to use just `_close_socket` method?

cannatag:
thanks!

PR #970, opened 2021-07-16
Update connection.rst to not have "s" undefined

Mindslave:
In the old version

```
from ldap3 import Server, Connection, SAFE_SYNC
server = Server('my_server')
conn = Connection(s, 'my_user', 'my_password', client_strategy=SAFE_SYNC, auto_bind=True)
status, result, response, _ = conn.search('o=test', '(objectclass=*)')  # usually you don't need the original request (4th element of the return tuple)
```

the variable "s" would be undefined, I'm pretty sure that this variable was supposed to reference the "server" created before. Thus I changed s to server.


cannatag:
thanks

PR #973, opened 2021-07-25
Made it safe to mix LDAPS and StartTLS servers in a pool.

zorn96:
Currently, if you use a restartable connection, it's only safe to use
entirely LDAPS or entirely StartTLS. If you have some servers that
would use StartTLS, but the first server selected by the Connection
uses LDAPS, then start_tls() will return False when called on the
connection as you can't use a StartLTS op on an LDAPS connection.

However, if that first server goes down later, the Restartable
connection may move to another server in the pool that CAN and
WANTS start_tls(). Previously, we'd only restart TLS if we'd
successfully started TLS. Alternatively, if you'd started TLS
previously, and then failed over to an LDAPS server, you'd
fail to restart TLS (since that's not valid on LDAPS) and this
would raise an exception.

So in summary, it wasn't safe to mix StartTLS and LDAPS servers in a
pool, because you would either lose TLS protection on a failover
or have the restart functionality break and raise an exception on a
failover.

This change makes it so that we won't try to restart TLS with StartTLS
if the connection's current server uses LDAPS (because we've already
reopened the connection and therefore TLS is working).
It also makes it so that we will try to StartTLS on restart if the
previous Server had started TLS successfully, OR if it used LDAPS.

This makes it safe to mix StartTLS and LDAPS servers in a pool.

### DIFF
--- a/ldap3/strategy/restartable.py
+++ b/ldap3/strategy/restartable.py
@@ -117,7 +117,10 @@ class RestartableStrategy(SyncStrategy):
         self._current_request = request
         self._current_controls = controls
         if not self._restart_tls:  # RFCs doesn't define how to stop tls once started
-            self._restart_tls = self.connection.tls_started
+            # to better support mixing LDAPS and StartTLS servers within a server pool, set TLS
+            # as something that should restart if we're using StartTLS or LDAPS right now, so
+            # that we don't protection if we failover across servers
+            self._restart_tls = self.connection.tls_started or self.connection.server.ssl
         if message_type == 'bindRequest':  # stores controls used in bind operation to be used again when restarting the connection
             self._last_bind_controls = controls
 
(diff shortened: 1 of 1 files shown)


zorn96:
@cannatag this seems like a pretty straightforward change, but wanted your thoughts on it.
I didn't address the Reusable strategy for connections because it seemed pretty convoluted and I didn't want to risk introducing a bug; also, I feel like I remember seeing something about it being deprecated.
I can go ahead and add it though if you feel like it'd be valuable

cannatag:
You’re right, the Reusable strategy is too hard to debug, so I flagged it as deprecated, also because is something you probably need on a server app and not in a client one.

I think it will be removed in the 3.0 version of ldap3, together with the support of python 2.7.

So, there is no need to work on it anymore.


Bye,
Giovanni

PR #977, opened 2021-08-03
D42-20011: fix how we lookup referrals

achmykhalo:


achmykhalo:
sorry, incorrect target

PR #1009, opened 2022-01-10
Update to MockBaseStrategy to add entries_from_dict in lieu of entries_from_json

andrewjroth:
This is a quick, non-breaking change to the MockBaseStrategy used for mocking the ldap3 connection.  This change adds a new method `entries_from_dict` using the definition from `entries_from_json` (without the JSON file open and decode) that allows entries to be loaded from a dictionary object.  The method `entries_from_json` was then updated to use the new method (same code as before) but keeps the file open and decode using `json.load` as it did before.  The end result does not change any usage of `entries_from_json`, but adds the new option to load data from an opened and decoded file.

Looking at unit tests, I wasn't able to find any existing test that covers the `entries_from_json` method, so no tests were added or modified to cover the new code.  Any tests written in the future that cover `entries_from_json` would also cover the new `entires_from_dict`.  These methods are only used during testing, so unit tests may not be required.

Thank you for considering this small PR!!

### DIFF
--- a/ldap3/strategy/mockBase.py
+++ b/ldap3/strategy/mockBase.py
@@ -262,14 +262,18 @@ class MockBaseStrategy(object):
     def entries_from_json(self, json_entry_file):
         target = open(json_entry_file, 'r')
         definition = json.load(target, object_hook=json_hook)
-        if 'entries' not in definition:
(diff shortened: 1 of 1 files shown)


cannatag:
Thanks!

PR #1042, opened 2022-07-08
Add Digest-MD5, NTLM and Kerberos encryption support

CravateRouge:
Add SASL Digest-MD5 encryption for confidentiality protection as described in [RFC2831](https://www.rfc-editor.org/rfc/rfc2831.html#section-2.4). 
Also add NTLM encryption for confidentiality protection as described in [MS-NLMP](https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-nlmp/b38c36ed-2804-4868-a9ff-8dd3182128e4).
And finally add SASL GSSAPI Kerberos encryption support.
Allows you to use LDAP without TLS for sensitive data exchange (e.g AD without LDAPS enabled, which is the default and you want to modify a user password).


zorn96:
this seems neat. @CravateRouge how have you tested this?

CravateRouge:
I made ldap requests on my AD, performing change passwords operation in simple LDAP. I used Wireshark to help me debug this mess.

ThePirateWhoSmellsOfSunflowers:
Very nice PR @CravateRouge, i've tested it and it works like a charm! It allows ldap3 to be used when LDAP Signing is enforced on Domain Controllers.

@cannatag have you planned to merge this one soon?

:sunflower:

quanah:
DIGEST-MD5 and NTLM are both historic and shouldn't be available at all.

vruello:
This PR looks great! I have tested the Kerberos encryption and it works like a charm.

Are there any plans to merge it in the near future?

maretodoric:
This should really be merged, works flawlessly :)

quanah:
I would disagree, DIGEST-MD5 and NTLM are both historic and should not be used.  Kerberos v5 support would be the only bit worth including.

quanah:
Better to support SCRAM mechanisms

maretodoric:
Perhaps, but until Windows keeps on deploying it by default, it should still be implemented in clients, as well.
I've deployed AD from AWS and by default ldaps is off and DIGEST-MD5 is supported/enabled.

Neustradamus:
I have seen this page: https://offsec.almond.consulting/ldap-authentication-in-active-directory-environments.html

I do not see SCRAM part, so I have created 2 tickets:
- https://github.com/cannatag/ldap3/issues/1106
- https://github.com/cannatag/ldap3/issues/1107

Thanks in advance.

cannatag:
Thanks!

PR #1062, opened 2022-10-05
Don't use auto_bind=True in documentation

lkiesow:
This patch switches from using the legacy option `auto_bind=True` to the equivalent new `auto_bind=AUTO_BIND_NO_TLS` in the documentation.


cannatag:
Thanks!

PR #1067, opened 2022-12-05
Set ciphers before validating the cert chain

AdrianVollmer:
This patch fixes a bug with TLS cipher settings not being honored.

If the cipher defaults are sufficiently strict (like they are in python3.10 by default), certificate chains with a certificate signed using MD5, for example, will be rejected with this error:

    wrap socket error: [SSL: CA_MD_TOO_WEAK] ca md too weak (_ssl.c:3874)

The cipher settings as specified by the user should be set before loading the certificate chain.


cannatag:
Thanks!

Issue #1051, opened 2022-08-23
Missing crypto dependency for MD4


exploide:
In https://github.com/cannatag/ldap3/blob/6c13304f03a7f66df40a43eccbe49ff75c2d6099/ldap3/utils/ntlm.py#L497--L501 the library attempts to calculate a MD4 hash.

This will only succeed with `hashlib` when the platform (OpenSSL) provides this algorithm, which becomes less common on recent systems.

As a fallback, the hash function is imported from the `Crypto` module, which comes from `pycryptodome` (or the older `pycrypto` which has the same interface).
But none of the libraries is mentioned as a dependency in the `requirements.txt` so it might happen that no library is available.

This results in such an error:

```
Traceback (most recent call last):
  File "/tmp/venv/lib/python3.10/site-packages/ldap3/utils/ntlm.py", line 500, in ntowf_v2
    from Crypto.Hash import MD4  # try with the Crypto library if present
ModuleNotFoundError: No module named 'Crypto'

During handling of the above exception, another exception occurred:

[...]
  File "/tmp/venv/lib/python3.10/site-packages/ldap3/utils/ntlm.py", line 497, in ntowf_v2
    password_digest = hashlib.new('MD4', self._password.encode('utf-16-le')).digest()
  File "/usr/lib64/python3.10/hashlib.py", line 166, in __hash_new
    return __get_builtin_constructor(name)(data)
  File "/usr/lib64/python3.10/hashlib.py", line 123, in __get_builtin_constructor
    raise ValueError('unsupported hash type ' + name)
ValueError: unsupported hash type MD4
```

So I propose adding `pycryptodome` to `requirements.txt`. On the other hand, since no backwards compatibility with `pycrypto` is necessary here, one could also adapt the code and depend on the recommended `pycryptodomex`.

At first, this issue is for documentation purposes, since I saw several people stumbling upon this. Furthermore, I could open a PR to fix this but I did not saw a lot of activity here recently. Is ldap3 still maintained and would a PR be welcome?

AzureIP:
Ok, so this is fixed but not released yet, correct? It will be released in 2.10.2rc2?
I experienced this error yesterday.

exploide:
> Ok, so this is fixed but not released yet, correct? It will be released in 2.10.2rc2?

Yes, the patch is contained in the 2.10.2rc2 pre-release but a stable release has not been made yet. This is tracked in #1169

0w15h3115:
[...]

PR #1076, opened 2023-01-24
Add missing crypto module to requirements.txt

nurfed1:
It seems that on newer systems, the MD4 function was removed from the hashlib module.  While I'm not sure if this is due to changes in python itself or in a third-party library such as openssl, this is an issue. 

The Ldap3 codebase uses the crypto module as a fallback for the MD4 function as referenced here: 
- https://github.com/cannatag/ldap3/blob/dev/ldap3/utils/ntlm.py#L500

However the pycryptodome module is not included in the ldap3 requirements files.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,6 @@
 pip
 pyasn1>=0.4.6
+pycryptodome
 setuptools
 sphinx
 sphinx-rtd-theme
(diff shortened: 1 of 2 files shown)


jonasagx:
Hi @cannatag do you see any blockers for this PR? LGTM and will help users.

jonasagx:
Closes https://github.com/cannatag/ldap3/issues/1051

PR #1089, opened 2023-05-10
Update tutorial_intro.rst usage of `auto_bind`

stevleibelt:
Update usage of `auto_bind` and replace legacy value of `True` with current hard coded `True`-Representation of `AUTO_BIND_NO_TLS`

See `connection.py` line 252ff

```python
            if auto_bind is False:  # compatibility with older version where auto_bind was a boolean
                self.auto_bind = AUTO_BIND_DEFAULT
            elif auto_bind is True:
                self.auto_bind = AUTO_BIND_NO_TLS
            else:
                self.auto_bind = auto_bind
```

### DIFF
--- a/docs/manual/source/tutorial_intro.rst
+++ b/docs/manual/source/tutorial_intro.rst
@@ -158,9 +158,9 @@ Let's start accessing the server with an anonymous bind::
 
 or shorter::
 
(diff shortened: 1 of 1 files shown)


cannatag:
thanks!

PR #1116, opened 2023-12-07
fix: execute method instead of return it

CaioCSdev:
Simple fix. I didn't manage to show a bug for this fix because I didn't manage to call `_split_ava` through "public" functions.

But it is clear that if it receives an ava starting with `=` and with `strip == True`, it will return something like `('', <built-in method strip of str object at 0xffffffffffff>)`, which is not what anyone wants.


cannatag:
Thanks!

PR #1132, opened 2024-03-20
Import MD4 from pycryptodomex package

### LINKED ISSUE #1038: Python 3.10 with ldap3.NTLM - unsupported hash type MD4
**Ubuntu 22.04 / Python 3.10.4 / ldap3==2.9 with authentication=ldap3.NTLM**

Exception:
Error: unsupported hash type MD4

### LINKED ISSUE #1051: Missing crypto dependency for MD4

exploide:
Now that the `pycryptodomex` package is listed as a requirement, we can just import `MD4` from the `Cryptodome` module to get rid of the annoying import problems.

They occurred when the system's OpenSSL does no longer enable MD4 and (and hence `hashlib` cannot use it) and the import from `Crypto` also fails because neither the legacy `pycrypto` package nor the `pycryptodome` package is installed. None of them was listed as a dependency. But now `pycryptodomex` is. :)

fixes #1051
fixes #1038

### DIFF
--- a/ldap3/utils/ntlm.py
+++ b/ldap3/utils/ntlm.py
@@ -512,15 +512,8 @@ class NtlmClient(object):
             # The specified password is an LM:NTLM hash
             password_digest = binascii.unhexlify(passparts[1])
         else:
-            try:
-                password_digest = hashlib.new('MD4', self._password.encode('utf-16-le')).digest()
-            except ValueError as e:
-                try:
-                    from Crypto.Hash import MD4  # try with the Crypto library if present
-                    password_digest = MD4.new(self._password.encode('utf-16-le')).digest()
-                except ImportError:
-                    raise e  # raise original exception
-
+            from Cryptodome.Hash import MD4
+            password_digest = MD4.new(self._password.encode('utf-16-le')).digest()
         return hmac.new(password_digest, (self.user_name.upper() + self.user_domain).encode('utf-16-le'), digestmod=hashlib.md5).digest()
 
     def _kxkey(self, response_key_nt, nt_proof_str):


cannatag:
thanks!

testworksau:
@cannatag there doesn't appear to be a release that includes this on pypi; is there a plan to do so soon?

Wissperwind:
Hi
Had the same issue and found this. When can we expect a release?

ilpssun:
@cannatag Could you release this version on pypi, please?  It keeps my project stuck on python 3.10 for the moment and I would like to upgrade to 3.12 soon because of other dependencies.

Alternatively, if you have no capacity to support the project, please just say so.  Then I'd look into other ldap libraries – although I'm not keen on doing so 😉

cannatag:
Hi, I’ll try to release a new version before the end of the year, during the Christmas season.Have fun,GiovanniIl giorno 5 dic 2024, alle ore 10:49, ilpssun ***@***.***> ha scritto:﻿
@cannatag Could you release this version on pypi, please?  It keeps my project stuck on python 3.10 for the moment and I would like to upgrade to 3.12 soon because of other dependencies.
Alternatively, if you have no capacity to support the project, please just say so.  Then I'd look into other ldap libraries – although I'm not keen on doing so 😉

jdmarques:
@cannatag any news on when will this be released?

ilpssun:
For everybody who is still struggling with this, #1051 gives an easy workaround: install `pycryptodome` explicitly, e.g., add it to your `requirements.txt`.

Issue #950, opened 2021-06-07
LDIF_LINE_LENGTH is ignored


AdrianVollmer:
Setting `LDIF_LINE_LENGTH` to a value that is not the default of 78 has no effect.

This is because `conf_ldif_line_length` is set at import time here: https://github.com/cannatag/ldap3/blob/dev/ldap3/protocol/rfc2849.py#L37

I propose moving it to line 234 into the function `operation_to_ldif`.

resposit:
@cannatag please apply this fix !

PR #1158, opened 2024-09-19
Fix up LDIF_LINE_LENGTH as per issue #950

plambrechtsen:
As per issue #950 the parameter `LDIF_LINE_LENGTH` doesn't work as it is global rather than in the `operation_to_ldif` funcion.


plambrechtsen:
This is a simple fix @cannatag that resolves an annoying bug I always need to patch when exporting LDIFs where I don't want to be RFC compliant and have each attribute on a single line.

plambrechtsen:
Thank you so much. 👍

PR #1161, opened 2024-10-25
Tls: add custom_ssl_context parameter

bgre:
Add a custom_ssl_context parameter to `Tls()` to pass a custom `ssl.SSLContext` object for TLS connections. This allows configurations that currently are not supported in the code, such as more recent ssl options like [ssl.SSLContext.verify_flags](https://docs.python.org/3.11/library/ssl.html#ssl.SSLContext.verify_flags).

One use case is to enable [partial certificate chain verification](https://docs.python.org/3.11/library/ssl.html#ssl.VERIFY_X509_PARTIAL_CHAIN) for pinning certificates in situations where a root certificate is not available.

### DIFF
--- a/ldap3/core/tls.py
+++ b/ldap3/core/tls.py
@@ -80,7 +80,32 @@ class Tls(object):
                  local_private_key_password=None,
                  ciphers=None,
                  sni=None,
-                 peer_certificate=None):
+                 peer_certificate=None,
+                 custom_ssl_context=None):
+        if custom_ssl_context is not None:
+            if not use_ssl_context:
+                if log_enabled(ERROR):
+                    log(ERROR, 'cannot use custom_ssl_context, SSLContext not available')
+                raise LDAPSSLNotSupportedError('cannot use custom_ssl_context, SSLContext not available')
+            if not is instance(custom_ssl_context, ssl.SSLContext):
+                if log_enabled(ERROR):
+                    log(ERROR, 'custom_ssl_context must be an ssl.SSLContext object')
+                raise LDAPSSLNotSupportedError('custom_ssl_context must be an ssl.SSLContext object')
+            if not (local_private_key_file is None and
+                 local_certificate_file is None and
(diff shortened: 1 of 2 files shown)


airtower-luna:
My I ask why this has been closed? Being able to provide the `ssl.SSLContext` to use would make TLS config a lot simpler *and* let people use all settings provided by the standard library. That's how I got here, ldap3 fails to connect to a server where `SSLContext.wrap_socket()` works fine (still investigating the difference, but I'd like to just use the context config that I know works).

With v3 (API break) it could even completely replace the internal TLS setup (which would simplify the code a lot).

sheep7:
I had the PR open as draft only (I guess that was a reason it was closed), because while it worked for my use-case, I didn't test it yet more thoroughly. Also, I discovered later that ldap3 creates TLS connections in other parts of the code (for referral connections) that my PR didn't cover. Solving this would require a refactor of my code, as custom TLS contexts cannot be re-used, so passing in a factory function might be a better solution.

airtower-luna:
> I had the PR open as draft only (I guess that was a reason it was closed), because while it worked for my use-case, I didn't test it yet more thoroughly. Also, I discovered later that ldap3 creates TLS connections in other parts of the code (for referral connections) that my PR didn't cover.

Thanks, I see!

> Solving this would require a refactor of my code, as custom TLS contexts cannot be re-used, so passing in a factory function might be a better solution.

Why shouldn't they be reused? The `ssl` module documentation [explicitly states they can](https://docs.python.org/3.13/library/ssl.html#ssl.SSLContext) (as long as they're not modified after first use):

> **Note:** [SSLContext](https://docs.python.org/3.13/library/ssl.html#ssl.SSLContext) is designed to be shared and used by multiple connections. Thus, it is thread-safe as long as it is not reconfigured after being used by a connection.

A factory function would *work*, but create a lot of unnecessary overhead if it's called for each new TLS connection.

PR #983, opened 2021-08-30
Updated tagMap and typeMap imports for pyasn1 0.5.0+

zorn96:
pyasn1 likes making breaking changes. We already do a lot
of versioned imports/logic, so this is just one more.

In 0.5.0, tagMap and typeMap are renamed to TAG_MAP and
TYPE_MAP

PR #1168, opened 2025-01-24
Updated tagMap and typeMap imports for pyasn1 0.5.0+

tjni:
Applying https://github.com/cannatag/ldap3/pull/983. In at least one instance, this deprecation warning caused confusion internally over whether it was the root cause of an unrelated error.


tjni:
Oops, I'm sorry. I opened this PR against the wrong repo.

PR #1139, opened 2024-04-12
fix requirements.txt

ThePirateWhoSmellsOfSunflowers:
Hello @cannatag ,

This PR is related to #1133 and should fix it.
Thanks,

:sunflower:

PR #1173, opened 2025-02-25
Require winkerberos only on Win32

david672orford:
Currently will not build on Linux because winkerberos is listed in requirements.txt and this package will not build on Linux. This PR adds a platform test to requirements.txt and requirements-dev.txt.


ThePirateWhoSmellsOfSunflowers:
I think it is duplicated with #1139

:sunflower:

